PR #2, opened 2020-03-03
Initial app architecture

Odrakir:
This is by no means complete, but it has enough stuff so everyone can review.

Do raise your concerns. Also, check the docs.

Once this is approved and merged we can start working with regular pull requests and issues.


App/App/AppDelegate.swift
```diff
@@ -0,0 +1,62 @@
+//
+//  AppDelegate.swift
+//  App
+//
+//  Created by Ricardo Sánchez Sotres on 19/02/2020.
+//  Copyright © 2020 Ricardo Sánchez Sotres. All rights reserved.
+//
```
amulware: I don't think we want these in our repo, do we?

Odrakir: These are a pain because as far as I know templates are global to Xcode, so there's no way to remove those comments just for this project. 

Anyway, I'll delete them.

amulware: Maybe we can set up some linting for this.

sofokli1: We can create our own templates files without the comments on top, not just that but we can create our own template files with editable boilerplate codes. This approach would take the guess work out of usual files like reducers, actions etc. (bonus we can have a template that creates multiple files at once (feature template) with things like `*Feature.Swift`, `*State.swift`, `*Action.swift`, `*Coordinator.swift`, where *=FeatureName). It would make things more uniform. We can have the templates in the project repo and have build script that copies those where they should be in each of our laptops so those would be always up to date.

What do you guys think?

ref https://thoughtbot.com/blog/creating-custom-xcode-templates#creating-a-template

Modules/Architecture/Architecture/Core/Reducer.swift
```diff
@@ -0,0 +1,72 @@
...
+    
+    public init(_ reduce: @escaping ReduceFunction<StateType, ActionType, EnvironmentType>)
+    {
+        self.reduce = reduce
+    }
+    
+    public func pullback<GlobalState, GlobalAction, GlobalEnvironment>(
+        state: WritableKeyPath<GlobalState, StateType>,
+        action: WritableKeyPath<GlobalAction, ActionType?>,
+        environment: KeyPath<GlobalEnvironment, EnvironmentType>
```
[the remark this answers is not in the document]
Odrakir: I'm ok with it as it is, because it's more concise and the intent is clear by the method name: pullback state, pullback action, pullback environment...

Modules/Architecture/Architecture/Core/BaseStore.swift
```diff
@@ -0,0 +1,84 @@
...
+    private var disposeBag: DisposeBag = DisposeBag()
+    
+    public var state: Driver<State>
+    public var _dispatch: (([Action]) -> Void)?
+            
+    private init(dispatch: @escaping ([Action]) -> Void, stateObservable: Driver<State>)
+    {
+        _dispatch = dispatch
+        state = stateObservable
+    }
```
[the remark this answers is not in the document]
Odrakir: It's called down from the `view` method

Modules/Assets/Assets/TextFiles/me.txt
```diff
@@ -0,0 +1 @@
+{"id":2750508,"api_token":"817238b5266da34673f758997c5141ce","email":"ricardo@toggl.com","fullname":"Ricardo","timezone":"Europe/Madrid","default_workspace_id":2051285,"beginning_of_week":1,"image_url":"https://assets.toggl.com/avatars/5becdf762fe18b2455e96656da8c7ea2.png","created_at":"2017-02-28T17:04:36.546618Z","updated_at":"2019-11-29T11:42:01.660987Z","openid_email":null,"openid_enabled":false,"country_id":208,"at":"2020-02-20T18:05:54.975429Z","intercom_hash":"ea06322cc5101195de603709ebfbe97ee92a6b0be9407783f045696118462c66"}
```
amulware: Just checking, should this personal data be here?

Modules/Models/Models/View/TimeEntryViewModel.swift
```diff
@@ -0,0 +1,111 @@
+//
+//  TimeEntryViewModel.swift
+//  Models
+//
+//  Created by Ricardo Sánchez Sotres on 03/03/2020.
+//  Copyright © 2020 Ricardo Sánchez Sotres. All rights reserved.
+//
```
amulware: More comments.

Odrakir: argh!

Modules/Onboarding/Onboarding/EmailLogin/Views/LoginViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+            .drive(userLabel.rx.text)
+            .disposed(by: disposeBag)
+        
+        self.navigationController?.presentationController?.delegate = self
+    }
+    
+    public override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        
+        store.dispatch(.emailEntered("ricardo@toggl.com"))
```
amulware: also this perhaps?

Modules/Assets/Assets/TextFiles/me.txt
Odrakir: It's completely innocuous, but yeah, we should replace it with something else. In terms of my privacy the harm is done (little harm), so let's do that in another PR.

PR #15, opened 2020-03-11
Use Cocoapods instead of SPM

### LINKED ISSUE #13: Use Cocoapods
The project is using Swift Package Manager for now, and though I like the simplicity of it, it's going to be problematic when including Firebase (and probably some other 3rd party dependencies too).

Using two package managers doesn't seem like a great idea, so we should migrate to Cocoapods, taking into consideration how the project frameworks are setup (a bunch of static frameworks with one dynamic framework)

Odrakir:
Closes #13 

I'm not certain about this at all, so I need some @toggl/ios-senseis with experience on Cocoapods to review, please.

The idea is to have this:

App
  ⎿ TogglTrack (dynamic)
         ⎿ Onboarding (static)
         ⎿ Timer (static)
         ⎿ Architecture (static)
         ⎿ ...         

I also took an idea from [here](https://medium.com/@acecilia/static-vs-dynamic-frameworks-in-swift-an-in-depth-analysis-ff61a77eec65) to make external dependencies static as well.

This does work (at least on my machine™️), but I'm not really sure what I'm doing. Please help!


dshr:
Hey! Started reading through the changes, and noticed one thing right off — the `Pods` dir should be in .gitignore IMO, if only just for cleanliness' sake. I'll post more things if they come up, but everything else looks good so far!

PR #16, opened 2020-03-11
Add Swiftlint

heytherewill:
This makes Swiftlint run on whenever we compile the app and closes #3 

Things we need to discuss:

- Whether we agree with the rules or not
- Whether we should use it from brew or from CocoaPods (pending #15)

The second commit is basically running `swiftlint autocorrect`. There still are some violations (like some force casts) that need to be solved manually. Please pull this branch locally, open it on XCode and see what the code looks like

PR #17, opened 2020-03-11
Add bitrise.yml

### LINKED ISSUE #4: Set up Continuous integration
## Description
We agreed on using Bitrise again for CI, since we are proficient at it and it works pretty well for our workflows

## Dependencies
#3 should be setup and merged before we can close this one

## Definition of done
- [ ] **IMPORTANT: The `bitrise.yml` file should be committed to our repo and all the secret keys needed that can't be public should be in bitrise. We should do it like so to make it simpler to review changes in bitrise and to avoid leaking tokens/password and the like**
- [ ] Checks run on every commit
- [ ] One bitrise flow checks that the app compiles just fine
- [ ] Other flow checks that the unit tests ran an passed
- [ ] The build must fail if any code is not formatted according to the linter rules
- [ ] Pull requests made against `master` cannot be merged unless all checks pass

heytherewill:
Closes #4

This adds the bitrise configuration file. It ensures that for every PR:

- Tests run
- The app compiles
- Swiftlint rules are enforced (needs #16)


Odrakir:
I added two steps that run first: base and dependencies. I'm doing this because we don't need to download all coredata dependencies to run the linter.

Odrakir:
Finally, I've also merged all workflows into one.

Odrakir:
@heytherewill I think we can remove the compile step, since there's no way to run the tests step without it compiling again. With these merged, for now, it doesn't make sense to do both one after the other.

I need you to remove the compile check from GH, though.

PR #28, opened 2020-03-16
Remove environment from reducers

Odrakir:
This is a proposal to follow Android approach and remove the environment from the reducers and instead injecting it when creating them.

Please, look for any shortcomings of this approach, are tests harder? Are dependencies more implicit?


Modules/Timer/Timer/TimeEntriesLog/TimeEntriesLogReducer.swift
```diff
@@ -32,11 +25,10 @@ func createTimeEntriesLogReducer(repository: Repository) -> Reducer<TimeEntriesL
             state.entities.timeEntries[timeEntry.id] = timeEntry
             return .empty
             
-        case .load:
-            if state.entities.loading.isLoaded {
-                return .empty
-            }
+        case .load where state.entities.loading.isLoaded:
```
Odrakir: Will we go over the complexity limit if we keep these two together? I'm asking because I feel this approach is more error prone as it depends on the order of the switch cases.

Modules/Timer/Timer/TimeEntriesLog/TimeEntriesLogReducer.swift
```diff
@@ -83,7 +88,8 @@ private func deleteTimeEntry(_ repository: Repository, timeEntryId: Int) -> Effe
     )
 }
 
-private func continueTimeEntry(_ repository: Repository, timeEntry: TimeEntry) -> Effect<TimeEntriesLogAction> {
+private func continueTimeEntry(_ repository: Repository, state: TimeEntriesLogState, timeEntryId: Int) -> Effect<TimeEntriesLogAction> {
+    guard let timeEntry = state.entities.timeEntries[timeEntryId] else { fatalError() }
```
Odrakir: As this is an effect you could maybe return a `setError` action instead of doing this? 🤔 I'm not really sure, maybe it's better to crash since this should not every happen.

heytherewill: Crashing is what we agreed upon prior, so let's stick to it

Modules/Timer/Timer/TimeEntriesLog/TimeEntriesLogReducer.swift
dshr: We don't, but it felt cleaner to me. Gonna revert this then.

Modules/Timer/Timer/TimeEntriesLog/TimeEntriesLogReducer.swift
dshr: Let's leave this as is it for now, and return to it if it ever causes trouble.

PR #32, opened 2020-03-17
Batch effects

### LINKED ISSUE #31: Improve the contract of our effects
## Description
Our effects can only ever emit a single action, but right now their interfaces allow for more. We need to fix this. 

## Definition of done
- [x] The effect type is remodeled to return 0 or 1 action instead of 0 or many as it is right now (IOW, A `Single` rather than an `Observable`

### LINKED ISSUE #35: Implement effect batching
## Description
In order to avoid rapid firing events from clogging up the UI, we need to give actions the ability to return an array of effects that can be partially applied and only emit the actual UI event once.

## Definition of done
- [x] Reducers return a list of effects rather than a single effect
- [x] These returned effects get applied all at once to the main state 
- [x] They do get applied partially in some intermediate state, so state consistency is achieved

Odrakir:
This PR changes 3 things:
- Makes effects contain a `Single<Action>` instead of subclassing observable. This limits effects to just emitting one action each.
- Makes reducers return an array of effects instead of just one effect. Since effects only emit one value, an array is how you emit multiple actions from one reducer.
- Refactors the store so all effects that any reducer return are dispatched in a batch.

The rest is refactoring the existing reducers. Specifically look at the `loadEntities` method in `timeEntriesLogReducer`

I also added a way to log when state changes happen. We will remove this way of logging in the future, but for now it comes handy.

Closes #31 
Closes #35


Modules/TogglTrack/TogglTrack/TogglTrack.swift
```diff
@@ -24,7 +24,11 @@ public class TogglTrack {
         
         appCoordinator = (appFeature.mainCoordinator(store: store) as? AppCoordinator)!
         router = Router(initialCoordinator: appCoordinator)
-                
+        
+        store.select({ $0 })
+            .drive(onNext: { _ in print(">>>>>>>>>>>>>>") })
```
heytherewill: Shouldn't this be done by the logging reducer? 🤔

Modules/Onboarding/Onboarding/EmailLogin/EmailLoginReducer.swift
```diff
@@ -11,30 +11,36 @@ func createEmailLoginReducer(api: UserAPI) -> Reducer<OnboardingState, EmailLogi
...
             state.route = AppRoute.onboarding
+            return []
             
         case let .emailEntered(email):
             state.email = email
+            return []
             
         case let .passwordEntered(password):
             state.password = password
+            return []
```
heytherewill: Maybe a `noEffect()` extensions would be nice to make this semantic (also because that's how it's done on the other platform)

Modules/Architecture/Architecture/Core/Store.swift
```diff
@@ -21,18 +21,20 @@ public final class Store<State, Action> {
         state = behaviorRelay.asDriver()
         
         internalDispatch = { actions in
+            
             var tempState = behaviorRelay.value
             var effects = [Effect<Action>]()
```
heytherewill: Why not `val effects = action.flatMap { $0.reduce(&tempState, action) }`?

Modules/TogglTrack/TogglTrack/TogglTrack.swift
heytherewill: Now that I think about it, that's impossible. We'd need to have a higher order store and not a higher order reducer... not sure we want that

Odrakir: I'm thinking we might want to implement better logging soon, one that can be switched on/off and one that let us choose the amount of logging and the destination (console, file). That logger could be a view into the store.

We should create an issue about that. But for now this is really helpful.

Modules/Onboarding/Onboarding/EmailLogin/EmailLoginReducer.swift
Odrakir: We used to have `.empty` for an empty effect. But this is not what it does anymore. This is a empty array, maybe an extension would make things more confusing? 🤔

heytherewill: Yes, we will

heytherewill: The good thing about having some `noEffect()` function is that you don't have to touch every reducer whenever you make a change to the return type, which happened a couple of times already on Droid :)

PR #33, opened 2020-03-17
Add StartEditAction.doneTapped action

### LINKED ISSUE #27: Add DoneButtonTapped action
## Description
As per what was discussed with @pdrj, we won't be saving the changes in the start time entry every time the user does something. Instead, we will keep the changes somewhere and commit them when the user emits a [`DoneButtonTapped`](https://www.notion.so/toggl/Start-time-entry-5783684cca064055b4512819087375bb?p=75aa87156f6e46859a5ce6a5711fea89) action. 

The action added in this issue should use gather all the intermediate changes and commit them to the database. 

## Definition of done
- [x] The action is added and handled by the reducer
- [ ] Tests cover that the action changes the state correctly

arturs105:
Closes #27 

Still a work on progress, as tests are not added yet.


Modules/Timer/Timer/StartEdit/StartEditReducer.swift
```diff
@@ -37,6 +37,38 @@ func createStartEditReducer(repository: Repository) -> Reducer<StartEditState, S
...
+            return .empty
+        }
+        state.timeEntryId = nil
+        state.description = ""
+        return updateTimeEntry(timeEntry, repository: repository)
+        
+    case let .timeEntryUpdated(timeEntry):
+        state.entities.timeEntries[timeEntry.id] = timeEntry
+        
+    case let .timeEntryAdded(timeEntry):
```
heytherewill: Both these cases could fall through

Modules/Timer/Timer/StartEdit/StartEditReducer.swift
```diff
@@ -37,6 +37,38 @@ func createStartEditReducer(repository: Repository) -> Reducer<StartEditState, S
...
+        let timeEntry = TimeEntry(
+            id: state.entities.timeEntries.count,
+            description: state.description,
+            start: Date(),
+            duration: -1,
+            billable: false,
+            workspaceId: defaultWorkspace
+        )
+        
+        state.timeEntryId = timeEntry.id
```
heytherewill: This id should be set on the `timeEntryStarted` method

Modules/Timer/Timer/StartEdit/StartEditReducer.swift
Odrakir: In swift switch statements don't fall through by default. But yeah, I think we should return in every statement to make things more explicit.

Modules/Timer/Timer/StartEdit/StartEditState.swift
```diff
@@ -6,4 +6,5 @@ public struct StartEditState {
     var user: Loadable<User>
     var entities: TimeLogEntities
     var description: String
+    var timeEntryId: Int?
```
Odrakir: I would use a more explicit name for this, in the line of `editingTimeEntryId` or similar.

Modules/Timer/Timer/StartEdit/StartEditReducer.swift
```diff
@@ -37,6 +37,38 @@ func createStartEditReducer(repository: Repository) -> Reducer<StartEditState, S
...
+        
+    case .doneTapped:
+        guard let timeEntryId = state.timeEntryId else {
+            return .empty
+        }
+        guard let timeEntry = state.entities.timeEntries[timeEntryId] else {
+            return .empty
+        }
+        state.timeEntryId = nil
+        state.description = ""
```
Odrakir: These two should be done in timeEntryUpdated, right? This way if there's an error we don't delete them and let the user try again.

Modules/Timer/Timer/StartEdit/StartEditReducer.swift
```diff
@@ -16,30 +16,57 @@ func createStartEditReducer(repository: TimeLogRepository, time: Time) -> Reduce
...
-        case let .timeEntryAdded(timeEntry):
+        case let .timeEntryUpdated(timeEntry):
             state.entities.timeEntries[timeEntry.id] = timeEntry
-            return []
             
-        case let .setError(error):
-            fatalError(error.description)
+        case let .timeEntryStarted(startedTimeEntry, stoppedTimeEntry):
+
+            state.editableTimeEntry = EditableTimeEntry.fromSingle(startedTimeEntry)
```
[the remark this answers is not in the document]
arturs105: I don't think it should. The `startTimeEntryButtonTapped` action calls `repository.startTimeEntry` and the repo then returns an observable which emits the started (and stopped) entry. That observable is then transformed to `timeEntryStarted` action, which is handled here. The editable entry isn't being set anywhere else.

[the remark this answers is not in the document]
heytherewill: I think Ricardo is correct. The only way this reducer is handling this action is when the user is hitting start. If he did so, the view is ready to be closed. You behavior *could* also be correct, thought; I think it boils down to a product decision. pinging @pdrj as the feature champion

Modules/Timer/Timer/StartEdit/Views/StartEditViewController.swift
```diff
@@ -41,6 +41,7 @@ public class StartEditViewController: UIViewController, Storyboarded, BottomShee
 
     public override func viewDidLoad() {
         super.viewDidLoad()
+
```
heytherewill, line 44: You can roll this back

Modules/Timer/Timer/StartEdit/StartEditReducer.swift
```diff
@@ -14,32 +14,49 @@ func createStartEditReducer(repository: TimeLogRepository, time: Time) -> Reduce
...
-        case let .setError(error):
-            fatalError(error.description)
+        case let .timeEntryStarted(startedTimeEntry, stoppedTimeEntry):
+            state.editableTimeEntry = nil
+            state.entities.timeEntries[startedTimeEntry.id] = startedTimeEntry
+            if let stoppedTimeEntry = stoppedTimeEntry {
+                state.entities.timeEntries[stoppedTimeEntry.id] = stoppedTimeEntry
+            }
         }
+        return []
```
Odrakir, line 49: Can you move this to every case that needs it instead of here just to be more explicit about it?

PR #34, opened 2020-03-17
Add a TimeService

### LINKED ISSUE #11: Add a TimeService
## Description
We need to add some interface capable of providing time when we need it. Unlike the current apps, which need something that ticks every second, this time service will have a single responsibility for starters: It will have a `now` method that returns the current time to allow for mocking.

## Definition of done
- [ ] A `TimeService` protocol is added
- [x] An implementation of the above is added
- [x] The service is injected in the adequate places
- [x] The service is used in the adequate places

dshr:
Closes #11

This PR adds a `Time` service, which then gets injected where needed: into `TimeEntriesLogReducer.swift` and `StartEditReducer.swift`. It also refactors a `TimeInterval` extension that relied on `Date` to not rely on it.


Modules/Timer/Timer/Common/TimerEnvironment.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+import Repository
+import Time
+
+public struct TimerEnvironment {
```
Odrakir: Is this used anywhere? With the new way to inject dependencies directly into reducersI think  we don't need this anymore.

dshr: Forgot about this file 🤦‍♂

dshr:
As for `OtherServices` — that makes sense, I'm gonna do that

heytherewill:
On droid we just added that to the `environment` module

Modules/OtherServices/OtherServicesTests/OtherServicesTests.swift
```diff
@@ -1,7 +1,15 @@
+//
+//  OtherServicesTests.swift
+//  OtherServicesTests
+//
+//  Created by Desislav Hristov on 17/03/2020.
+//  Copyright © 2020 Toggl OU. All rights reserved.
+//
+
```
Odrakir: :fire:

Modules/OtherServices/OtherServices/OtherServices.h
```diff
@@ -0,0 +1,19 @@
+//
+//  OtherServices.h
+//  OtherServices
+//
+//  Created by Desislav Hristov on 17/03/2020.
+//  Copyright © 2020 Toggl OU. All rights reserved.
+//
```
Odrakir: :fire:

PR #39, opened 2020-03-18
Time entry log grouping

### LINKED ISSUE #21: Add logic for grouping time entries
## Description
We need to add the logic that groups similar time entries, similar to what [we have in the current app](https://github.com/toggl/mobileapp/blob/83418effb77e5957a68679c0ab2c4fe49ec36512/Toggl.Core.Tests/UI/Transformations/TimeEntriesFlatenningTests.cs). Since there's no switch to enable/disable grouping, we should leave grouping on by default until we start implementing the settings view

## Definition of done
- [ ] Similar time entries are turned into a group
- [ ] Tests ensure the grouping logic is correct

Odrakir:
This adds grouping to the time entry log.

Closes #21 
(I see now I didn't need to do the UI, but still it's much easier to test it with it, so...)

Two things left to do
- Add tests. I'll add an issue for it once the base code for tests is written
- Move all the datasource code to its own class out of the view controller. Basically because that view controller is getting too big. I'll add another issue for this, doing it here seemed like too much already.

BTW tableview animations seem broken in the simulator when deleting cells. It works fine on device, though. I haven't been able to find out why.


App/App.xcodeproj/project.pbxproj
```diff
@@ -474,7 +474,7 @@
 			baseConfigurationReference = 94C5C24F604EAAF62DD4C477 /* Pods-App.release.xcconfig */;
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
-				CODE_SIGN_IDENTITY = "iPhone Distribution";
+				CODE_SIGN_IDENTITY = "iPhone Developer";
```
arturs105: Is this change necessary?

Modules/Timer/Timer/TimeEntriesLog/TimeEntriesLogAction.swift
```diff
@@ -13,6 +13,9 @@ public enum TimeEntriesLogAction {
     
     case timeEntryDeleted(Int)
     case timeEntryAdded(TimeEntry)
+
+    case toggleTimeEntryGroupTapped(Int)
```
arturs105: The parameter type doesn't match the notion page (notion has `List<long>`). I see how this can work with a single id but we should keep the code and the doc consistent with each other. Updating the doc according to this code should be fine (unless @heytherewill disagrees).

Modules/Timer/Timer/TimeEntriesLog/ViewModels/TimeLogCellViewModel.swift
```diff
@@ -0,0 +1,116 @@
+import UIKit
+import Models
+
+enum TimeLogCellViewModel: Equatable {
+    
+    public static func == (lhs: TimeLogCellViewModel, rhs: TimeLogCellViewModel) -> Bool {
+        switch (lhs, rhs) {
+        case (let .singleEntry(lentry, linGroup), let .singleEntry(rentry, rinGroup)):
```
arturs105: `linGroup` -> `lInGroup`, or even better: `leftInGroup`. Same for `rinGroup`

Modules/Timer/Timer/TimeEntriesLog/ViewModels/TimeLogCellViewModel.swift
```diff
@@ -0,0 +1,116 @@
...
+            return lentry == rentry && linGroup == rinGroup
+        case (let .groupedEntriesHeader(lentries, lopen), let .groupedEntriesHeader(rentries, ropen)):
+            return lentries == rentries && lopen == ropen
```
arturs105: `lEntries` and `lOpen`

Modules/Timer/Timer/TimeEntriesLog/Views/TimeEntriesLogViewController.swift
```diff
@@ -23,7 +23,10 @@ public class TimeEntriesLogViewController: UIViewController, Storyboarded {
 
     public override func viewDidLoad() {
         super.viewDidLoad()
+        
         tableView.rowHeight = 72
+        tableView.sectionHeaderHeight = 44
+        tableView.register(TimeLogHeaderView.self, forHeaderFooterViewReuseIdentifier: "TimeLogHeader")
```
arturs105: Let's make the string literals into constants. Especially since they are repeated more than once.

Modules/Timer/Timer/TimeEntriesLog/Views/TimeEntriesLogViewController.swift
```diff
@@ -78,35 +94,51 @@ extension TimeEntriesLogViewController: UITableViewDelegate {
...
+            } else {
+                self.store.dispatch(TimeEntriesLogAction.timeEntrySwiped(.right, timeLogCellViewModel.mainEntryId))
+            }
         }
         action.backgroundColor = .green
         return UISwipeActionsConfiguration(actions: [action])
     }
 
     public func tableView(_ tableView: UITableView, trailingSwipeActionsConfigurationForRowAt indexPath: IndexPath) -> UISwipeActionsConfiguration? {
         let action = UIContextualAction(style: .destructive, title: "Delete") { _, _, _ in
```
arturs105: Not a part of this PR, but are we not using string resources from day one for easier translation ?

Modules/Timer/Timer/TimeEntriesLog/Views/TimeLogHeaderView.swift
```diff
@@ -0,0 +1,48 @@
+//
+//  TimeLogHeaderView.swift
+//  Timer
+//
+//  Created by Ricardo Sánchez Sotres on 16/03/2020.
+//  Copyright © 2020 Ricardo Sánchez Sotres. All rights reserved.
+//
+
```
arturs105: 🔥

Modules/Timer/Timer/TimeEntriesLog/Views/TimeLogHeaderView.swift
```diff
@@ -0,0 +1,48 @@
...
+        contentView.addSubview(descriptionLabel)
+        contentView.addSubview(durationLabel)
+        descriptionLabel.translatesAutoresizingMaskIntoConstraints = false
+        durationLabel.translatesAutoresizingMaskIntoConstraints = false
+        descriptionLabel.centerYAnchor.constraint(equalTo: contentView.centerYAnchor).isActive = true
+        descriptionLabel.leadingAnchor.constraint(equalTo: contentView.leadingAnchor, constant: 10).isActive = true
+        durationLabel.centerYAnchor.constraint(equalTo: contentView.centerYAnchor).isActive = true
+        durationLabel.trailingAnchor.constraint(equalTo: contentView.trailingAnchor, constant: -10).isActive = true
+        descriptionLabel.trailingAnchor.constraint(equalTo: durationLabel.leadingAnchor, constant: 20).isActive = true
+        durationLabel.widthAnchor.constraint(equalToConstant: 80).isActive = true
```
arturs105: Can't this be done with a xib or a storyboard?

App/App.xcodeproj/project.pbxproj
Odrakir: Nope

Odrakir:
Superseded by #49 and a subsequent one with the UI

PR #41, opened 2020-03-18
Add |> and <| operators for easier reducer chaining

### LINKED ISSUE #30: Improve the way we chain high order reducers
Currently the way we chain high-order is not scalable beautifully - it get's too lisp-y too fast. Example: `someHighOrderReducer(anotherHighOrderReducer(appReducer))`. Let's make it more beautiful with some DSL.

I had an idea for something like:
```
appReducer
    .andThen(someOtherReducer)
    .andThen(theNextReducer)
```
Though it seems awfully similar to `reducer.combine` function.

@heytherewill had some other ideas as well.

dshr:
This PR introduces the `|>` and `<|` operators, which are standard in other FP languages. There was some doubts about it, so we have an alternative approach in #42.
This allows us to go from this:
``` swift
reducer: logging(checkFeatureAvailability(createAppReducer(environment: AppEnvironment())))
```
to the much more sane
``` swift
reducer: logging
    <| checkFeatureAvailability
    <| createAppReducer(environment: AppEnvironment())
```

Closes #30


heytherewill:
As I said in Slack, let's move on with the operators

PR #42, opened 2020-03-18
Add a way to chain higher-order reducers

dshr:
This PR introduces a different way to chain reducers, without relying on custom operators (ref #41). Here we go from this:
``` swift
reducer: logging(checkFeatureAvailability(createAppReducer(environment: AppEnvironment())))
```
to this:
``` swift
reducer: chain([logging, checkFeatureAvailability],
    reducer: createAppReducer(environment: AppEnvironment()))
```


heytherewill:
Superseded in favor of #41

PR #43, opened 2020-03-19
Update docs

Odrakir:
Updates the docs with the latest changes to the architecture.

(it also deletes some commented code we missed)


README.md
```diff
@@ -20,9 +20,9 @@ This is a high level overview of the different parts of the architecture.
...
-- **Environment** This will hold all the dependencies our reducers need, so basically everything in the infrastructure layer. Examples could be the DataBase, the API, a timer service, a location service... This is passed into the store on construction and the store takes care of injecting it into the reducers.
+- **Reducers** Reducers are pure functions that take the state and an action and produce a new state. Simple as that. In our case they also take the environment (just the part they need) and optionally result in an array of Effects that will asynchronously dispatch further actions. All business logic should reside in them.
+- **Effects** As mentioned, Reducers optionally produce these after handling an action. They are just singles, as in observables that emit just one more action. Those observables are subscribed to in the store and the emitted action will be dispatched again. These are use as a way to handle asynchronous actions. All the effects emitted from a reducer will be batched, meaning the state change will only be emitted once all actions are dispatched.
```
heytherewill:
```suggestion
- **Effects** As mentioned, Reducers optionally produce these after handling an action. They are just singles, as in observables that emit just one or no action. Those observables are subscribed to in the store and the emitted action will be dispatched again. These are use as a way to handle asynchronous actions. All the effects emitted from a reducer will be batched, meaning the state change will only be emitted once all actions are dispatched.
```

README.md
```diff
@@ -109,45 +108,51 @@ public func batch(_ actions: [Action])
...
+(inout StateType, ActionType) -> [Effect<ActionType>]
 ```
 
 The idea is they take the state and an action and modify the state depending on the action and its payload.
 
-Reducers also take an `Environment` which will contain all the dependencies they need to do their tasks (Repository, API, Keychain...).
+Reducers are usually constructed with part of the `Environment` which will contain all the dependencies they need to do their tasks (Repository, API, Keychain...).
 
-In order to dispatch actions asynchronously we will use `Effects`. Reducers return an `Effect` which is an observable of actions. The store subscribes to those effects and dispatches whatever actions it emits.
+In order to dispatch actions asynchronously we will use `Effects`. Reducers return an array of `Effects` which is an single of action. The store subscribes to those effects and dispatches whatever actions they emit.
```
heytherewill:
```suggestion
In order to dispatch actions asynchronously we will use `Effects`. Reducers return an array of `Effects` which is a `Single<Action>`. The store subscribes to those effects and dispatches whatever action they emit, if any.
```

PR #44, opened 2020-03-20
Extract loading

Odrakir:
This extracts the loading of entities to its own reducer and outside of the log reducer.

I'm doing this in order to simplify state and make it equatable son tests are simpler.


Modules/TogglTrack/TogglTrack/Loading/LoadingReducer.swift
```diff
@@ -0,0 +1,67 @@
...
+        repository.getTasks().map(LoadingAction.setTasks),
+        repository.getTags().map(LoadingAction.setTags),
+        Single.just(LoadingAction.loadingFinished)
+    ]
+        .map { $0.toEffect() }
+}
+
+private func arrayToDict<EntityType: Entity>(entities: [EntityType]) -> [Int: EntityType] {
+    return entities.reduce([:], { acc, entity in
+        var acc = acc
```
arturs105: What's the purpose of this line?

Odrakir: acc is an immutable, we copy it into a mutable variable with the same name (yeah, you can do that in Swift) so we can mutate it.

Issue #36, opened 2020-03-17
Add proper time entry persistency


heytherewill:
## Description
We need to persist time entries in disk properly. After discussing with the sync team, we should start doing it.

## Definition of done
- [ ] CoreData is properly configured to persist time entries
- [ ] The repository is changed to use the actual database interface
- [ ] The implementation, as discussed with @Odrakir, should follow the [principles of the Android implementation](https://github.com/toggl/android/blob/master/database/src/main/java/com/toggl/database/dao/TimeEntryDao.kt). This is to make this layer of app similar so that the sync team can act on it without much hurdle when context switching

heytherewill:
Closed by #47

PR #47, opened 2020-03-20
Add proper time entry persistency

dshr:
This PR will add CoreData and all necessary code to persist TimeEntries, thus closing #36.


Modules/Database/Database/Models/TimeEntry.swift
```diff
@@ -0,0 +1,134 @@
...
+        context.delete(managedTimeEntry)
+    }
+    
+    // Delete by id
+    public func delete(id: Int) throws {
+        let context = database.persistentContainer.viewContext
+        let request = generateFetchByIdRequest(Int64(id))
+        let managedTimeEntry = try context.fetch(request).first!
+        context.delete(managedTimeEntry)
+    }
```
Odrakir: Why do we need these both?

Modules/Database/Database/Models/TimeEntry.swift
```diff
@@ -0,0 +1,134 @@
...
+    // Update
+    public func updateAll(timeEntries: [TimeEntry]) throws {
+        try timeEntries.forEach { try update(timeEntry: $0) }
+    }
+
+    // Delete
+    public func delete(timeEntry: TimeEntry) throws {
+        let context = database.persistentContainer.viewContext
+        let request = generateFetchByIdRequest(Int64(timeEntry.id))
+        let managedTimeEntry = try context.fetch(request).first!
```
Odrakir: If there are no results the app will crash, maybe we should throw or respond in some better way

Modules/Repository/Repository/Repository.swift
```diff
@@ -86,25 +89,44 @@ extension Repository: TimeLogRepository {
...
     public func startTimeEntry(timeEntry: TimeEntry) -> Single<(started: TimeEntry, stopped: TimeEntry?)> {
-        
         var stoppedTimeEntry: TimeEntry?
-        if let runningEntryIndex = timeEntries.firstIndex(where: { $0.duration == 0}) {
-            stoppedTimeEntry = timeEntries[runningEntryIndex]
-            stoppedTimeEntry!.duration = timeEntry.start.timeIntervalSince(stoppedTimeEntry!.start)
-            timeEntries[runningEntryIndex] = stoppedTimeEntry!
+        do {
+            let timeEntries = try database.timeEntries.getAll()
+            if let runningEntryIndex = timeEntries.firstIndex(where: { $0.duration == 0}) {
```
Odrakir: In the DB query you are testing for `duration == nil`, we should unify this

Modules/Repository/Repository/Repository.swift
```diff
@@ -86,25 +89,44 @@ extension Repository: TimeLogRepository {
...
+                stoppedTimeEntry = timeEntries[runningEntryIndex]
+                stoppedTimeEntry!.duration = timeEntry.start.timeIntervalSince(stoppedTimeEntry!.start)
+                try database.timeEntries.update(timeEntry: stoppedTimeEntry!)
+            }
+
+            var newTimeEntry = timeEntry
+            newTimeEntry.id = (timeEntries.map({ $0.id }).max() ?? 0) + 1
```
Odrakir: This obviously is not how ids should be calculated. Maybe leave a comment here? We need local ids and server ids (which will be returned upon sync)

Modules/Repository/Repository/Repository.swift
```diff
@@ -86,25 +89,44 @@ extension Repository: TimeLogRepository {
...
+            _ = try database.timeEntries.insert(timeEntry: newTimeEntry)
+            return Single.just((started: newTimeEntry, stopped: stoppedTimeEntry))
+        } catch let error {
+            print(error)
```
Odrakir: All these should probably be returning errors.

Modules/Database/Database/Models/TimeEntry.swift
```diff
@@ -0,0 +1,134 @@
...
+    
+    // Query("SELECT * FROM TimeEntry WHERE NOT isDeleted")
+    public func getAll() throws -> [TimeEntry] {
+        let request = NSFetchRequest<ManagedTimeEntry>(entityName: String(describing: ManagedTimeEntry.self))
+        request.predicate = NSPredicate(format: "isServerDeleted == false")
+        return try executeFetchRequest(request)
+    }
+
+    // Query("SELECT * FROM TimeEntry WHERE NOT isDeleted AND duration is null")
+    public func getAllRunning() throws -> [TimeEntry] {
```
Odrakir: 🤔 Are we ever going to need this one?

Modules/Database/Database/Models/TimeEntry.swift
```diff
@@ -0,0 +1,96 @@
+import Foundation
+import Models
+import CoreData
+
+extension TimeEntry {
+    fileprivate static func fromManagedObject(managedTimeEntry: ManagedTimeEntry) -> TimeEntry {
+        return TimeEntry(
+            id: Int(managedTimeEntry.id),
+            description: managedTimeEntry.textDescription!,
+            start: managedTimeEntry.start!,
```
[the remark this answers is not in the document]
dshr: Apparently that "Optional" means a different thing, so we can't make these non-optional, not without explicitly editing the generated `NSManagedObject` subclasses. Source: https://stackoverflow.com/a/52033823

Modules/Database/Database/Models/TimeEntry.swift
dshr: Good point. The first one was created following the spec, the second one was for use with the current delete method from the repository. I'll still make the first one call the second, since it only makes sense.

Modules/Repository/Repository/Repository.swift
dshr: This is not my code, but something I got during the mergeback. I'll add the comment nonetheless.

Modules/Repository/Repository/Repository.swift
dshr: So we mark all Repository methods as `throws`?

Modules/Database/Database/Models/TimeEntry.swift
dshr: I was following the spec, so no idea.

Modules/Database/Database/Models/TimeEntry.swift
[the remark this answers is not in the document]
Odrakir: Ok, but still, all those properties are optional in the DB model schema and some of them at least shouldn't be. 

Also, I think it might be a good idea to create the NSManagedObject subclass manually so we can add our own code, since we are going to be using it in the whole repo layer...

Modules/Repository/Repository/Repository.swift
Odrakir: Yeah, that's my code, but most of what's laying around was there as a temporary solution to build the prototype fast. Better than adding the comment do create a UUID, which is what the repo guys will need.

Modules/Repository/Repository/Repository.swift
Odrakir: `Single.error`

Modules/Repository/Repository/Repository.swift
heytherewill: For now, ids should simply be generated sequentially in the database. No need to worry about the UUIDs for the time being

Modules/Database/Database/Models/TimeEntry.swift
```diff
@@ -80,6 +80,8 @@ public struct TimeEntriesDatabase {
             into: context)
             as? ManagedTimeEntry else { throw SaveFailedError() }
         managedTimeEntry = TimeEntry.toManagedObject(timeEntry: timeEntry, managedTimeEntry: managedTimeEntry)
+        let lastId = try getAll().last?.id ?? -1
```
heytherewill: This looks like an overkill. Is there no way of having automatic database handled incrementing/unique ids on CoreData?

Modules/Database/Database/Database.swift
```diff
@@ -0,0 +1,36 @@
...
+    public init() { }
+}
+
+public class Database {
+    private let identifier = "com.toggl.aurora.Database"
+    private let model = "Database"
+    
+    lazy public var timeEntries = TimeEntriesDatabase(database: self)
+    
+    public init() {
```
heytherewill: Is the empty constructor needed?

Modules/Database/Database/Models/TimeEntry.swift
```diff
@@ -0,0 +1,133 @@
...
+    private let database: Database
+    
+    public init(database: Database) {
+        self.database = database
+    }
+    
+    // Query("SELECT * FROM TimeEntry WHERE NOT isDeleted")
+    public func getAll() throws -> [TimeEntry] {
+        let request = NSFetchRequest<ManagedTimeEntry>(entityName: String(describing: ManagedTimeEntry.self))
+        request.predicate = NSPredicate(format: "isServerDeleted == false")
```
heytherewill: Why is the property called `isServerDeleted`? There is no such property in the returned JSON models; the `isDeleted` property is used for our own soft deleting + syncing purposes

Modules/Database/DatabaseTests/DatabaseTests.swift
```diff
@@ -0,0 +1,26 @@
+import XCTest
+@testable import Database
+
+class DatabaseTests: XCTestCase {
```
heytherewill: Is this file needed?

Modules/Repository/Repository/Repository.swift
```diff
@@ -86,25 +89,46 @@ extension Repository: TimeLogRepository {
...
+        do {
+            try database.timeEntries.delete(id: timeEntryId)
+            return Single.just(())
+        } catch let error {
+            print(error)
+            return Single.error(error)
+        }
+    }
+    
+    private func fetchTimeEntriesFromAPI() -> Single<[TimeEntry]> {
```
heytherewill: It makes no sense to fetch from the API, even a mocked one. If the db is empty, we should simply return no entries.

Modules/Database/Database/Models/TimeEntry.swift
dshr: `isDeleted` is a reserved property of NSManagedObject.

Modules/Database/DatabaseTests/DatabaseTests.swift
dshr: It was auto-generated.

Modules/Database/Database/Models/TimeEntry.swift
dshr: There's [NSManagedObjectID](https://developer.apple.com/documentation/coredata/nsmanagedobjectid), but it's its own thing, and doesn't conform to `Int64`, so injecting it into our `Models.TimeEntry` is pointless.

Modules/Database/Database/Database.swift
dshr: It is needed indeed, in Swift classes need an explicit public constructor. And we can't have it as a struct bc the `TimeEntriesDatabase` is referencing the `Database`.

Modules/Database/Database/Models/TimeEntry.swift
dshr: Right, re-read your comment — CoreData has it's own soft deletion, so we can use that. Gonna delete this then.

Modules/Database/Database/Models/TimeEntry.swift
Odrakir: Couldn't this fail? Shouldn't we be using max instead?

But anyway, @simonrozsival mentioned generating a UUID (v4), so maybe we have to change the ID type and it can't be an Int anymore...

Modules/Database/DatabaseTests/DatabaseTests.swift
Odrakir: Tests will be added here. It's generated when you say your project will have tests. It can be aded afterwards, but I think we can live with it for now.

Modules/Repository/Repository/Repository.swift
```diff
@@ -52,12 +54,13 @@ extension Repository: TimeLogRepository {
...
     
     public func getTimeEntries() -> Single<[TimeEntry]> {
-        if timeEntries.isEmpty {
-            return api.loadEntries()
-                .do(onSuccess: { self.timeEntries = $0 })
+        do {
+            let timeEntries = try database.timeEntries.getAll()
+            return Single.just(timeEntries)
+        } catch let error {
+            print(error)
```
Odrakir: :fire:

Modules/Repository/Repository/Repository.swift
```diff
@@ -86,25 +89,34 @@ extension Repository: TimeLogRepository {
...
+                stoppedTimeEntry = runningEntry
+                stoppedTimeEntry!.duration = timeEntry.start.timeIntervalSince(stoppedTimeEntry!.start)
+                try database.timeEntries.update(timeEntry: stoppedTimeEntry!)
+            }
+
+            let newTimeEntry = timeEntry
+            _ = try database.timeEntries.insert(timeEntry: newTimeEntry)
+            return Single.just((started: newTimeEntry, stopped: stoppedTimeEntry))
+        } catch let error {
+            print(error)
```
Odrakir: :fire:

Modules/Repository/Repository/Repository.swift
```diff
@@ -86,25 +89,34 @@ extension Repository: TimeLogRepository {
...
     }
     
     public func deleteTimeEntry(timeEntryId: Int) -> Single<Void> {
-        timeEntries = timeEntries.filter { $0.id != timeEntryId }
-        return Single.just(())
+        do {
+            try database.timeEntries.delete(id: timeEntryId)
+            return Single.just(())
+        } catch let error {
+            print(error)
```
Odrakir: :fire:

Modules/Database/Database/Models/TimeEntry.swift
```diff
@@ -0,0 +1,132 @@
+import Foundation
+import Models
+import CoreData
+
+extension TimeEntry {
```
Odrakir: As discussed in Slack, I think it makes sense to handle the NSManagedObject subclass ourselves and turn off CoreData code generation. This way we can avoid using TimeEntry here, we can take/return `ManagedTimeEntry` objects (or a protocol so the repo doesn't know about DB details) with all the extra data the repo cares about but the business logic does not (like `isServerDeleted`)

```swift
final class ManagedTimeEntry: NSManagedObject {
    @NSManaged private(set) var id: Int
    @NSManaged private(set) var textDescription: String
    @NSManaged private(set) var start: Date
    @NSManaged private(set) var duration: NSNumber?
    @NSManaged private(set) var billable: Bool
}
```

Odrakir: Just manually add that, disable the code generation for the entity and maybe add a protocol so the repo can send/receive this.

Modules/Repository/Repository/Repository.swift
dshr: Remove the print statement?

Modules/Repository/Repository/Models/TimeEntry.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+import Models
+import Database
+
+extension TimeEntry {
```
dshr: Not quite sure about the naming of this file, plus maybe we can extract the Time Entry-specific methods here too, so the Repository.swift is just a base? What do you think, @Odrakir?

Modules/Database/Database/Models/TimeEntriesDatabase.swift
```diff
@@ -0,0 +1,125 @@
+import Foundation
+import Models
+import CoreData
+
+public protocol TimeEntryDTO {
```
heytherewill: Is a protocol with a single implementation really the best way to go?

Modules/Database/Database/Models/TimeEntriesDatabase.swift
```diff
@@ -0,0 +1,125 @@
...
+        managedTimeEntry.id = Int64(timeEntry.id)
+        managedTimeEntry.textDescription = timeEntry.description
+        managedTimeEntry.start = timeEntry.start
+        managedTimeEntry.duration = timeEntry.duration
+        managedTimeEntry.billable = timeEntry.billable
+        return managedTimeEntry
+    }
+}
+
+public struct TimeEntriesDatabase {
```
heytherewill: Wouldn't it be better to have this behind a protocol so that, when testing repository behavior, we don't need to know the implementation detail of this?

dshr: Good point, I'll add the appropriate protocol.

Modules/Database/Database/Models/TimeEntriesDatabase.swift
dshr: This is done so we don't expose the CoreData ManagedObject to the Repo, the other option being doing the transform to TimeEntry in the Database, but we agreed to do that in the Repo.

Modules/Database/Database/Models/TimeEntriesDatabase.swift
dshr: ✅

Modules/Repository/Repository/Models/TimeEntry.swift
Odrakir: The idea is to not import Models at all in the DB module. The Repository will be the one in charge of transforming TimeEntry to TimeEntryDAO (I prefer this name rather than TimeEntryDTO, as DTOs are used for other stuff).

So all DB methods return AND take TimeEntryDAO and not TimeEntry

Modules/Database/Database/Models/TimeEntriesDatabase.swift
Odrakir: There's an alternative that requieres less methods to mock, which is to create `DBResources` that describe the operation we want to do, so the DB in the end only exposes one method. We did this for Superday: https://github.com/toggl/superday/tree/develop/teferi/Services/Implementations/New%20Persistency/CoreData

I'm not saying we have to do this now or that we have to do it at all.

Modules/Database/Database/Models/TimeEntriesDatabase.swift
Odrakir: We have DB models, Repo models and app models. But in the case of the repo and the DB, as they'll share a lot of stuff we'll use a common protocol that hides the DB objects.

And then the repo transforms to and from app models when communicating with reducers.

Modules/Database/Database.xcdatamodeld/Database.xcdatamodel/contents
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="16119" systemVersion="19D76" minimumToolsVersion="Automatic" sourceLanguage="Swift" userDefinedModelVersionIdentifier="">
+    <entity name="ManagedTimeEntry" representedClassName=".ManagedTimeEntry" syncable="YES">
+        <attribute name="billable" optional="YES" attributeType="Boolean" usesScalarValueType="YES"/>
+        <attribute name="duration" optional="YES" attributeType="Double" defaultValueString="0.0" usesScalarValueType="YES"/>
+        <attribute name="id" optional="YES" attributeType="Integer 64" defaultValueString="0" usesScalarValueType="YES"/>
+        <attribute name="start" attributeType="Date" usesScalarValueType="NO"/>
+        <attribute name="textDescription" attributeType="String"/>
+        <attribute name="workspaceId" optional="YES" attributeType="Integer 64" defaultValueString="0" usesScalarValueType="YES"/>
```
dshr: Shouldn’t this be non-optional?

PR #49, opened 2020-03-23
Time entry log grouping logic

### LINKED ISSUE #21: Add logic for grouping time entries

Odrakir:
I'm splitting #39 in two. This one takes care of the logic only. 

Closes #21 

It includes the tests as we had them in the previous app (minus the date formatting stuff which will be tested elsewhere when included)

The `toggleTimeEntryGroupTapped` action takes a groupId. Ids should be type safe, but I'm leaving that for another PR as it's a big change.


Modules/Timer/Timer/TimeEntriesLog/ViewModels/TimeEntryViewModel.swift
```diff
@@ -24,34 +25,32 @@ public struct TimeEntryViewModel: Equatable {
...
+        self.groupId = timeEntry.groupId
         self.description = timeEntry.description.isEmpty ? "No description" : timeEntry.description
         self.projectTaskClient = getProjectTaskClient(from: project, task: task, client: client)
         self.billable = timeEntry.billable
                 
         self.start = timeEntry.start
         self.duration = timeEntry.duration >= 0 ? timeEntry.duration : nil
         if let duration = duration {
-            self.durationString = duration.formatted
+            self.durationString = duration.formattedDuration(ending: Date())
```
heytherewill: Shouldn't we be avoiding using raw dates like that?

Modules/Timer/Timer/TimeEntriesLog/ViewModels/DayViewModel.swift
```diff
@@ -1,15 +1,30 @@
...
+        day = timeLogCells.first!.start.ignoreTimeComponents()
         dayString = day.toDayString()
-        self.timeEntries = timeEntries
+        self.timeLogCells = timeLogCells
+        
+        durationString = self.timeLogCells
+                .filter({ !$0.isInGroup })
+                .map({ $0.duration ?? 0 })
+                .reduce(0, +)
+                .formattedDuration(ending: Date())
```
heytherewill: Same comment about dates

Modules/Timer/Timer/TimeEntriesLog/TimeEntriesSelector.swift
```diff
@@ -1,10 +1,69 @@
...
         .map(DayViewModel.init)
         .sorted(by: { $0.day > $1.day })
 }
+
+let groupTimeEntries: (Set<Int>) -> ([TimeEntryViewModel]) -> [TimeLogCellViewModel] = { expandedGroups in
+
+    return { timeEntryViewModels in
+        return timeEntryViewModels
+            .grouped { $0.groupId }
+            .sorted(by: { $0.first!.start > $1.first!.start })
```
heytherewill: Why not sort before grouping for a better looking syntax?

Modules/Utils/Utils/Extensions/TimeInterval+Extensions.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+
+extension TimeInterval {
+    public func formattedDuration(ending endingDate: Date) -> String {
```
heytherewill: This function should also be unit tested

Odrakir:
I'll create an issue to deal with date formatting, because it's not part of this PR. About functional testing, this is still missing the UI PR. Coming up next.

Modules/Timer/Timer/TimeEntriesLog/TimeEntriesSelector.swift
Odrakir: The intention was to sort the groups, not the individual entries. But anyway, this made me realise we were not sorting properly and we were not testing for the sorting either. I fixed that.

Modules/Timer/Timer/TimeEntriesLog/ViewModels/TimeEntryViewModel.swift
Odrakir: As I said, let's do dates in another PR

Modules/Timer/Timer/TimeEntriesLog/ViewModels/DayViewModel.swift
Odrakir: As I said, let's do dates in another PR

Modules/Utils/Utils/Extensions/TimeInterval+Extensions.swift
Odrakir: As I said, let's do dates in another PR

Modules/Timer/TimerTests/TestUtils/TimeEntry+Extensions.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+import Models
+
+extension TimeEntry {
+    static func with(id: Int? = nil, description: String = "", start: Date? = nil, duration: TimeInterval? = nil, workspaceId: Int = 0) -> TimeEntry {
+        let timeEntryId = id ?? UUID().hashValue
```
dshr: Wait, I thought we weren't doing UUIDs yet?

Odrakir: It's really not important. I just need a random Int for the tests.

dshr: Cool-cool

PR #56, opened 2020-03-24
Implement EditableTimeEntry

### LINKED ISSUE #55: Implement EditableTimeEntry
## Description
`EditableTimeEntry` functionality should be modelled after this PR: https://github.com/toggl/android/pull/79
## Definition of done
- [ ] The `EditableTimeEntry` type is added and is used in the `StartEditState` and `TimerState`

arturs105:
Closes #55 

The implementation is copied from toggl/android#79


Modules/Models/Models/Workspace.swift
```diff
@@ -5,10 +5,25 @@ public struct Workspace: Codable, Entity, Equatable {
...
+    public init(id: Int, name: String, admin: Bool, features: [WorkspaceFeature]) {
+        self.id = id
+        self.name = name
+        self.admin = admin
+        self.features = features
+    }
+}
+
+extension Workspace {
+    var irPro: Bool {
```
heytherewill:
```suggestion
    var isPro: Bool {
```

Modules/Models/Models/WorkspaceFeature.swift
```diff
@@ -0,0 +1,24 @@
+import Foundation
+
+public enum WorkspaceFeature : Int, Codable {
```
Odrakir: I don't know where does this come from and why do we need it. I think this shouldn't be part of this PR, right?

Modules/Models/Models/WorkspaceFeature.swift
```diff
@@ -0,0 +1,24 @@
...
+    case Free = 0
+    case Pro = 13
+    @available(*, deprecated, message: "Use specific granular features instead")
```
Odrakir: Why are we adding code that's already deprecated?!

Modules/Repository/Repository/Repository.swift
```diff
@@ -10,7 +11,7 @@ public protocol TimeLogRepository {
     func getProjects() -> Single<[Project]>
     func getTasks() -> Single<[Task]>
     func getTags() -> Single<[Tag]>
-    func startTimeEntry(timeEntry: TimeEntry) -> Single<(started: TimeEntry, stopped: TimeEntry?)>
+    func startTimeEntry(workspaceId: Int, description: String) -> Single<(started: TimeEntry, stopped: TimeEntry?)>
```
Odrakir: Pretty soon will be adding other parameters here, doesn't it make sense to send a struct instead? If you don't want to use `TimeEntry` because it has an id, maybe we can create a `NewTimeEntry` struct or use the `EditableTimeEntry` moving the ids property out of it.

Modules/Repository/Repository/Repository.swift
```diff
@@ -87,20 +90,29 @@ extension Repository: TimeLogRepository {
...
+    public func startTimeEntry(workspaceId: Int, description: String) -> Single<(started: TimeEntry, stopped: TimeEntry?)> {
         
-        var stoppedTimeEntry: TimeEntry?
-        if let runningEntryIndex = timeEntries.firstIndex(where: { $0.duration == 0}) {
-            stoppedTimeEntry = timeEntries[runningEntryIndex]
-            stoppedTimeEntry!.duration = timeEntry.start.timeIntervalSince(stoppedTimeEntry!.start)
-            timeEntries[runningEntryIndex] = stoppedTimeEntry!
+        let stoppedTimeEntry = stopRunningTimeEntry()
+        let startedTimeEntry = TimeEntry(
+            id: (timeEntries.map({ $0.id }).max() ?? 0) + 1,
```
Odrakir: Des is also touching this code. This should be a `UUID`

Modules/Repository/Repository/Repository.swift
```diff
@@ -87,20 +90,29 @@ extension Repository: TimeLogRepository {
...
+            description: description,
+            start: time.now(),
+            duration: -1,
```
Odrakir: Not sure if this is supposed to be negative, 0 or `nil`... can you verify this?

Modules/Timer/Timer/Common/EditableTimeEntry.swift
```diff
@@ -0,0 +1,46 @@
...
+import Models
+
+public struct EditableTimeEntry : Equatable {
+    public var ids: [Int]
+    public var workspaceId: Int
+    public var description: String
+    public var billable: Bool
+    
+    private init(
+        ids: [Int],
```
Odrakir: As I mentioned above, maybe having a struct to gather all edited data (`TimeEntryDTO` for example) and move the ids array out of it could allow us to use this struct when sending data to the repository in `startTimeEntry`

Modules/Timer/Timer/StartEdit/StartEditReducer.swift
```diff
@@ -7,29 +7,22 @@ import OtherServices
 
 func createStartEditReducer(repository: TimeLogRepository, time: Time) -> Reducer<StartEditState, StartEditAction> {
     return Reducer {state, action in
-
+
```
Odrakir: We will probably need another action to set the initial data when editing a TE. Not sure if it belongs to this PR, just mentioning it here.

Modules/Models/Models/Workspace.swift
arturs105: This seems like a brain fart because `is` in Latvian is `ir` 😄

Modules/Models/Models/WorkspaceFeature.swift
arturs105: It was added in the droid repo as well. @heytherewill might know the reason

heytherewill: Copied from the Xamarin repo. We might still receive this from API calls, but if we ever have to handle these in code, we should get notified and then discuss with backend how to proceed.

Modules/Repository/Repository/Repository.swift
heytherewill: for `startTimeEntry` it'd be good to have a `NewTimeEntry` struct that doesn't have a duration property, as described in this issue: https://github.com/toggl/mobileapp/issues/6452

Modules/Repository/Repository/Repository.swift
arturs105: If you check out the `master` branch, open up `StartEditReducer.swift` and go to `case .startTapped`, you can see the duration being set to -1. That's the reason i went with the same approach.

However, i think, it would be better if duration was nullable and was set to null for running time entries.

Modules/Timer/Timer/StartEdit/StartEditReducer.swift
arturs105: Good point, but definitely shouldn't be a part of this PR.

Modules/Models/Models/WorkspaceFeature.swift
Odrakir: Ok to this specific case. But please let's don't copy the other's platform code without questioning it, this is the perfect recipe for bugs.

Modules/Repository/Repository/Repository.swift
Odrakir: Ok, I'll add an issue

Modules/Models/Models/WorkspaceFeature.swift
heytherewill: Don't worry, I'm not doing it :)

Modules/Repository/Repository/TimeEntryDto.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+import Models
+
+public struct TimeEntryDto {
```
heytherewill: This name is not specific enough. Call it `StartTimeEntryDTO` or `NewRunningTimeEntryDTO`, but make sure that the name makes it explicit about its usage to prevent people from abusing it.

Modules/Models/Models/Workspace.swift
```diff
@@ -5,25 +5,16 @@ public struct Workspace: Codable, Entity, Equatable {
...
-    public init(id: Int, name: String, admin: Bool, features: [WorkspaceFeature]) {
+    public init(id: Int, name: String, admin: Bool) {
         self.id = id
         self.name = name
         self.admin = admin
-        self.features = features
-    }
-}
-
-extension Workspace {
```
heytherewill: Why was this deleted? Are you adding it only in the PR that checks it?

heytherewill: I'm not arguing for it to be brought back, I'm merely questioning

arturs105: Yeah, it was removed because it's not related to this PR. Gonna add it when implementing the `toggleBillableTapped` action, since it's needed there.

Modules/TogglTrack/TogglTrack/App/AppState.swift
```diff
@@ -12,7 +12,8 @@ public struct AppState {
     
     public var localLoadingState: LocalLoadingState = LocalLoadingState()
     public var localOnboardingState: LocalOnboardingState = LocalOnboardingState()
-    public var localTimerState: LocalTimerState = LocalTimerState()
+    public var localTimerState: LocalTimerState = LocalTimerState(
+        defaultWorkspace: Workspace(id: 1, name: "Auto created workspace", admin: true))
```
Odrakir: This should never be done like this. I guess this  is temporal, but still I think it makes sense to move the default workspace to either the user property, its own property in the app state or as a selector from user and workspaces.

I think it's worth it to make that change in this PR because building the `LocalTimerState` like this is very wrong.

So, I would add one of these, which will be set on the `LoadingReducer`:
- `defaultWorkspace` property in `TimeLogEntities`
- `defaultWorkspace` property in `User`
- `defaultWorkspace` property in `AppState`

@heytherewill thoughts?

Modules/Timer/Timer/Common/EditableTimeEntry.swift
```diff
@@ -0,0 +1,53 @@
+import Foundation
+import Models
+import Repository
+
+public struct EditableTimeEntry: Equatable {
```
Odrakir: Since we now have the `TimeEntryDTO` we don't need this struct anymore. We could have a `TimeEntryDTO` and an `[Int]` for the ids in the `StartEditState` instead. I think it's much cleaner. @heytherewill ?

arturs105: 🤔 That would work, yes. But my thinking was that the `StartTimeEntryDto` is immutable (notice the `let`s) and has a single purpose of passing data necessary to start a time entry. We'd later have other single purpose DTOs for other functionality. 

On the other hand, `EditableTimeEntry` is, well, editable, hence the `var` properties and is used for the single purpose of editing.

**tl;dr**: I think we should keep both because of single responsibility principle

Modules/TogglTrack/TogglTrack/App/AppState.swift
arturs105: Yeah, i copied this from the droid repo and had the same thought that this smells and is temporal so i should have questioned it, but i didn't in order to proceed at high velocity.

That being said, i support this change and we should pick one of the three options you're offering. Not sure which one though 🤔

Modules/Timer/Timer/Common/EditableTimeEntry.swift
Odrakir: Immutability is not that important for a DTO, because once you pass it around every method has its own copy, it's ok if they mutate it, the caller won't see the changes. I still think reusing the DTO is cleaner, the single responsibility principle doesn't apply here since it's all the same responsibility of holding the TE's data, but I would like to hear @heytherewill opinion.

Modules/TogglTrack/TogglTrack/App/AppState.swift
Odrakir: Any would do, this is easy to change in the future. 🤔 . If you want me to decide... let's put it into the `TimeLogEntities` struct.

arturs105: `TimeLogEntities ` struct it is then

Modules/Timer/Timer/Common/EditableTimeEntry.swift
heytherewill: >  I still think reusing the DTO is cleaner
This caused a lot of problems in the old app, because then you can't ensure constraints in compile time. When you start a time entry, it can't have a duration and when supporting premium features there might be even more of those, so ensuring that during compile time is much better IMO. Also, it doesn't need an implementation of the protocol and doesn't give you false expectations about the usage of `startTimeEntry` (e.g.: Passing a DTO that has a duration and expecting it to create a time entry that's not running).

Bottomline: I'd use specific DTOs, like was agreed for the old apps but never implemented, rather than reusing an interface, which was a mistake made in the old app. And no, I'm not blindly saying we should avoid the fears of the old app; I really think this is the best way to design the api for the repo.

heytherewill: I just noticed that here you are commenting on the `EditableTimeEntry` and not on the start time entry DTO. I think `EditableTimeEntry` is still the way to go, though, because it's a concrete type with everything that we need rather than two properties that aren't immediately connected 🤔

Odrakir: I'm not sure I understand. I'm not saying `TimeEntryDTO` should be a protocol, it's a struct. Basically the same as `EditableTimeEntry`, but keeping the ids out of it in another property.

Modules/TogglTrack/TogglTrack/App/AppState.swift
juanlaube: I'd say it should be in `User` since it's a property of the user in the business logic. However in some particular cases the default workspace in the user's profile is `null` (probably an error in the API or something, I don't think that's intended) and we need to make a decision, in the old app we decided to pick the workspace with the lowest id if I remember correctly.

Modules/Timer/Timer/StartEdit/StartEditReducer.swift
```diff
@@ -42,6 +37,9 @@ func createStartEditReducer(repository: TimeLogRepository, time: Time) -> Reduce
     }
 }
 
-func startTimeEntry(_ timeEntry: TimeEntry, repository: TimeLogRepository) -> Effect<StartEditAction> {
-    return Effect.empty
+func startTimeEntry(_ timeEntry: StartTimeEntryDto, repository: TimeLogRepository) -> Effect<StartEditAction> {
+    return repository
+        .startTimeEntry(timeEntry)
+        .toEffect(map: { (started, _) in StartEditAction.timeEntryAdded(started) },
```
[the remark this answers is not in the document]
arturs105: No, not yet. That action is added in #33 and that PR depends on this one.

arturs105:
Also: i'd rather add the `StartEditReducer ` tests in #33, since that PR changes the actions a bit.

PR #57, opened 2020-03-25
More time log reducer tests

### LINKED ISSUE #24: Add tests to the timer module
## Description
We need to catch up on Unit testing the Timer module. We should also consider including some solution for property based testing, like [SwiftCheck](https://github.com/typelift/SwiftCheck)

## Definition of done
- [ ] A property based testing library is included
- [ ] All existing actions of the timer module have relevant tests that cover all possible paths

Odrakir:
Closes #24 

This adds tests for all (current in this branch) user actions in the TimeEntriesLog feature.

The actions returned from effects do not have their own tests, as all of those are tested in flows initiated by the user.

One thing I'm not happy with is that we have most logic in the repository (due to wanting atomic operations). Which results in the reducer code being too simple and this tests not testing much. For example, in `testContinueTappedStoppingPreviousFlow` I set the duration of the stopped time entry, but it doesn't really matter as I'm mocking the repository and it'll return what ever I say.

I'm not sure if there's a better way, though. I'm just pointing out my concerns.


App/App.xcodeproj/project.pbxproj
```diff
@@ -520,7 +519,6 @@
 				CODE_SIGN_STYLE = Automatic;
 				DEVELOPMENT_TEAM = B227VTMZ94;
 				INFOPLIST_FILE = AppTests/Info.plist;
-				IPHONEOS_DEPLOYMENT_TARGET = 13.2;
```
heytherewill: What's with the changes in this file?

Odrakir: Somehow we forgot to change that deployment version, tests won't run if we don't change that.

PR #76, opened 2020-03-30
Switch to Int64 for all entity ids

### LINKED ISSUE #75: Migrate all entity ids to Int64

arturs105:
Closes #75

This changes (almost) all entity ids to use `Int64`. The only ids that are still `Int` are time entry group ids, because those are created locally by hashing properties of grouped entries and the hashing function returns `Int`.


juanlaube:
Looks good!

PR #79, opened 2020-03-31
🎨 Reflect client name in time entry cell

### LINKED ISSUE #54: Display client names in the main log
## Description
The name of the client should appear next to the project name on both simple and grouped time entries

## Dependencies
Depends on #53 

## Definition of done
- [ ] The client name, when present, is displayed next to the project name in the main log

juanlaube:
This reflects the client name in the time entry cells.
It uses an attributed string that concatenates the project, task and client and applies some attributes just like the old app.

Closes #54


Modules/Timer/Timer/TimeEntriesLog/ViewModels/TimeEntryViewModel.swift
```diff
@@ -3,62 +3,92 @@ import Models
...
+
+    public let start: Date
+    public let duration: TimeInterval?
+    public let durationString: String?
+    public let end: Date?
+    public let isRunning: Bool
+
+    public let descriptionColor: UIColor
+    public let projectColor: UIColor
+
```
heytherewill: What's with all these changes?

Modules/Timer/Timer/TimeEntriesLog/ViewModels/TimeEntryViewModel.swift
```diff
@@ -3,62 +3,92 @@ import Models
...
     public let tags: [Tag]?
-
```
heytherewill: All these empty space changes make it really hard to review; could you consider creating a separate PR for 💄 ?

Modules/Timer/Timer/TimeEntriesLog/ViewModels/TimeEntryViewModel.swift
```diff
@@ -3,62 +3,92 @@ import Models
...
         client: Client? = nil,
         task: Task? = nil,
         tags: [Tag]? = nil) {
-                
+
         self.id = timeEntry.id
         self.groupId = timeEntry.groupId
         self.description = timeEntry.description.isEmpty ? "No description" : timeEntry.description
-        self.projectTaskClient = getProjectTaskClient(from: project, task: task, client: client)
+        self.projectTaskClient = ProjectTaskClientFormatter.stringFrom(project: project, task: task, client: client)
```
heytherewill: Why do you need a simple string as well?

Modules/Timer/Timer/TimeEntriesLog/ViewModels/TimeEntryViewModel.swift
juanlaube: These properties are not being mutated, should be constants instead of variables.

Modules/Timer/Timer/TimeEntriesLog/ViewModels/TimeEntryViewModel.swift
juanlaube: Deleted, it's not actually needed.

heytherewill:
Please don't merge until @Odrakir has reviewed

Modules/Timer/Timer/TimeEntriesLog/ViewModels/TimeEntryViewModel.swift
```diff
@@ -37,28 +37,42 @@ public struct TimeEntryViewModel: Equatable {
...
     }
-}
 
-func getProjectTaskClient(from project: Project?, task: Task?, client: Client?) -> String {
-    var value = ""
-    if let project = project { value.append(project.name) }
-    if let task = task { value.append(": " + task.name) }
-    if let client = client { value.append(" · " + client.name) }
-    return value
+    public struct ProjectTaskClientFormatter {
```
Odrakir, line 55: Why does this need to be a struct? It could be a free function or an extension on `NSAttributedString`. Also, it will be used in other places, maybe its worth it to move it to `Utils`

juanlaube: I'm not a fan of free functions unless they really make sense, for now I moved it back to a free function, we can extract it back to that struct in the future when/if we need it.

PR #82, opened 2020-04-01
✨ Keep ids of deleted time entries in timer state

### LINKED ISSUE #63: Add a property to keep track of time entries that are pending deletion
## Description
As discussed [here](https://github.com/toggl/mobile-discussions/issues/58), we will need a property to temporarily store the entries that are soon-to-be deleted in order to properly update the UI and allow for the undo functionality

## Definition of done
- [ ] The `entriesPendingDeletion: HashSet<Long>` property is added to `TimerLocalState`
- [ ]  The `entriesPendingDeletion: HashSet<Long>` property is added to `TimeEntriesLogState`
- [ ] The property is considered in the TimeEntriesLog selector and any entries in that set are immediately filtered out from the UI
- [ ] Tests are added to the selector to ensure its behavior is correct

juanlaube:
This keeps track of the ids of the deleted time entries in the timer state.
Closes #63


Modules/Timer/TimerTests/Selectors/TimeEntriesSelectorTests.swift
```diff
@@ -12,21 +12,29 @@ class TimeEntriesSelectorTests: XCTestCase {
```
Odrakir, line 12: Are you sure it's a good idea to modify every single test instead of adding new ones to test just this?

juanlaube: Good point, I will revert those changes add them again as new tests

Odrakir: Am I mistaken or those changes were not reverted?

Odrakir:
@juanlaube I see you still changed all previous tests. Please explain.

Modules/Timer/TimerTests/Selectors/TimeEntriesSelectorTests.swift
juanlaube: Done!

PR #83, opened 2020-04-02
🎨 Add Snackbar view

### LINKED ISSUE #81: Implement a Snackbar mechanism
## Description
Since iOS does not have default snackbars, we should implement our own. As discussed [here](https://github.com/toggl/mobile-discussions/issues/59), most of the team agrees on making something similar to what we already have on Xamarin (but make it more Swift-y). Let's do so and ask for @Odrakir's blessing.

I think a good change we can do is make sure that the snackbar is bound to a specific kind of action? Something like: `Snackbar<TimeEntriesLogAction>.from(store, .undoButtonTapped).show()` sounds reasonable enough. 

## Definition of done
- [ ] Some method that displays a snackbar is added to the codebase
- [ ] Calling said method makes the snackbar appear for 5 seconds and then disappear
- [ ] The snackbar method has an undo button that is wired to emitting some action

juanlaube:
This adds the Snackbar view to the Utils module. The API is similar to Android's Snackbar for consistency.
I also added a new module, UIUtils, that depends on Utils and Architecture. This module adds an extension method on the Snackbar to use it with a store.

Closes #81


Modules/Utils/Utils/UI/Snackbar.swift
```diff
@@ -0,0 +1,136 @@
...
+public final class Snackbar: UIView {
+
+    private let height: CGFloat = 64
+    private let horizontalPadding: CGFloat = 20
+    private let verticalPadding: CGFloat = 10
+    private let verticalPaddingInWindow: CGFloat = 40
+    private let animationDuration: TimeInterval = 0.3
+
+    public enum Duration {
+        case short
```
heytherewill, line 17: Given that our snackbars only have one possible duration, I don't think this Duration enum is even needed

Modules/Utils/Utils/UI/Snackbar.swift
```diff
@@ -0,0 +1,136 @@
...
+        
+        var duration: TimeInterval {
+            switch self {
+            case .short: return 5
```
heytherewill, line 21: Do we have a constants file? I'd move this there if possible

Modules/Utils/Utils/UI/Snackbar.swift
```diff
@@ -0,0 +1,136 @@
...
+import UIKit
+
+public protocol SnackbarHost {
+    var snackbarHostBottomAnchor: NSLayoutYAxisAnchor { get }
+}
+
+public final class Snackbar: UIView {
+
+    private let height: CGFloat = 64
+    private let horizontalPadding: CGFloat = 20
```
heytherewill, line 11: The problem of having paddings here is that if we ever decide to change the padding of the views displaying the toasts, we'd need to remember and update this file, which is not ideal

Modules/Utils/Utils/UI/Snackbar.swift
juanlaube: We don't have one, but I made it a constant in the Snackbar class.

Modules/Utils/Utils/UI/Snackbar.swift
juanlaube: Paddings are gone 🔥 
I used the view controller's layout margins.

Modules/UIUtils/UIUtils/UIUtils.h
```diff
@@ -0,0 +1,19 @@
+//
+//  UIUtils.h
+//  UIUtils
+//
+//  Created by Juan Laube on 4/2/20.
+//  Copyright © 2020 Toggl Track. All rights reserved.
+//
+
```
heytherewill, line 8: 🔥 ?

Modules/UIUtils/UIUtilsTests/UIUtilsTests.swift
```diff
@@ -0,0 +1,34 @@
+//
+//  UIUtilsTests.swift
+//  UIUtilsTests
+//
+//  Created by Juan Laube on 4/2/20.
+//  Copyright © 2020 Toggl Track. All rights reserved.
+//
+
```
heytherewill, line 8: 🔥

Modules/UIUtils/UIUtils/UIUtils.h
juanlaube: Stupid Xcode 🤦‍♂

Modules/UIUtils/UIUtilsTests/UIUtilsTests.swift
juanlaube: Same

Odrakir:
By the way, why is the checksum in the Podfile.lock changing?

Issue #69, opened 2020-03-30
Add running time entry feature


Odrakir:
## Description
Add the feature that will display the card with either the running time entry with the stop button or the description textfield with the play button at the bottom of the timeline. This includes the view controller with the preliminary UI, the coordinator, the feature state and an empty reducer and actions.

This feature needs access to the currently running time entry to show its details or the empty start card if its null

## Dependencies
None

## Definition of done
- [x] The whole file structure is created inside the Timer module
- [x] The UI for both the running time entry and the small start card is created
- [x] The UI uses a selector for the current running time entry to adapt its layout
- [x] The empty reducer is created and combined with the rest into the app reducer

sofokli1:
@Odrakir should the description text input allow for multiple lines of text (thus needing to resize the view based on text amount)?

Odrakir:
I wouldn't resize the view, it should take the same space regardless of the content. Currently we have the moving label, not sure if this is what we want for the new app. Either talk with Alicja or do whatever you feel because design is not final right now.

PR #84, opened 2020-04-02
Preliminary UI for start/edit view

### LINKED ISSUE #38: Add preliminary UI for the start time entry card
## Description
Following the layouts [shown on invision](https://toggl.invisionapp.com/d/main#/console/19353169/407785375/preview), let's implement a very crude and barebones UI so that we can start hooking things up. Each person that develops a feature can further polish the part they are touching to make it look final.

## Definition of done
- [ ] The components of the start time entry card are present, even if they look ugly/off
- [ ] The card can swipe up and down
- [ ] *The components need not to be bound to the running time entry*

Odrakir:
Closes #38 

There's a bunch of stuff still missing, but this PR is getting big and it already adds a bottom sheet, with a table view for the start/edit view fields, an accessory view for the buttons above the keyboard. It handles panning for both the bottom sheet and the table view and changing between them (more or less) seamlessly.

I think this one can be reviewed and then we can create issues to:
- Bind this to be shown/hidden through state changes
- Bind with state
- Add the specific fields in the start/edit view (there's a lot of work here)
- Detect changes and present alert view when dismissing with changes pending
- Implement the real designs

Each these will probably involve multiple issues.

**Reviews** There are a lot of small commits, don't be scared by the number of them.


Odrakir:
The running entry view will come in #69. 

And tes, the dismissing is missing confirmation, as I say in the PR description. It should be done in another issue.

PR #85, opened 2020-04-02
Add running timeEntry Feature

### LINKED ISSUE #69: Add running time entry feature

sofokli1:
Closes #69

It is not complete 100% but wanted some feedback before I spend more time on this. Still need to update the elapsed time label and the description project task etc labels and colors.


juanlaube:
I didn't see anything I don't like, however I also reviewed #84 and both are doing changes in the BottomSheet.

Odrakir:
Yeah, I should have clarified this. @sofokli1 the running time entry view and the start edit view used to be a single view, but they are not any more. So the bottom sheet is used in the start edit view only (which will appear on top of this one). The running time entry view doesn't need the bottom sheet as it'll not be draggable.

Modules/Assets/Assets/Storyboards/Timer.storyboard
```diff
@@ -4,6 +4,7 @@
     <dependencies>
         <deployment identifier="iOS"/>
         <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="16086"/>
+        <capability name="Named colors" minToolsVersion="9.0"/>
```
Odrakir, line 7: What's this?

Modules/Timer/Timer/Common/SimpleBottomSheet.swift
```diff
@@ -0,0 +1,93 @@
+import UIKit
+import Utils
+import Assets
+
+class SimpleBottomSheet: UIViewController {
+
+    private var bottomConstraint: NSLayoutConstraint!
+    private var heightConstraint: NSLayoutConstraint!
```
Odrakir, line 8: You don't need these two

Modules/Timer/Timer/Common/SimpleBottomSheet.swift
```diff
@@ -0,0 +1,93 @@
...
+
+    public var stops: [CGFloat] = [77]
```
Odrakir, line 10: :fire:

Modules/Timer/Timer/Common/SimpleBottomSheet.swift
```diff
@@ -0,0 +1,93 @@
...
+        NotificationCenter.default.addObserver(
+            self,
+            selector: #selector(keyboardWillShow),
+            name: UIResponder.keyboardWillShowNotification,
+            object: nil)
+        NotificationCenter.default.addObserver(
+            self,
+            selector: #selector(keyboardWillHide),
+            name: UIResponder.keyboardWillHideNotification,
+            object: nil)
```
Odrakir, line 36: This view doesn't need to react to the keyboard, because when the TF is tapped we will show the start/edit view. So 🔥 (as a matter fact the TF doesn't even need to be a TF, it just needs to look like one)

Modules/Timer/Timer/Common/SimpleBottomSheet.swift
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    @objc
+    func keyboardWillHide(notification: NSNotification) {
+
+        UIView.animate(withDuration: 0.3) {
+            self.parent?.additionalSafeAreaInsets.bottom = 0
+            self.parent?.view.layoutIfNeeded()
+        }
+    }
```
Odrakir, line 92: :fire:

Modules/Timer/Timer/Common/Views/TimerViewController.swift
```diff
@@ -19,13 +27,27 @@ public class TimerViewController: UIViewController {
...
     }
+
+    private func updateRunningTimeEntryButtomSheet(isRunningTimeEntry: Bool) {
+        if isRunningTimeEntry {
+            uninstall(bottomSheet)
+            install(runningTimeEntryBottomSheet, customConstraints: true)
+        } else {
+            uninstall(runningTimeEntryBottomSheet)
+            install(bottomSheet, customConstraints: true)
+        }
```
Odrakir, line 51: 🤔 What's this for? When we stop the current running TE we should change the runningTEView to show the textfield and the play button, and when we start it we should show the running TE details. There's no need to install or uninstall those views.

Modules/Timer/Timer/RunningTimrEntry/RunningTimeEntryAction.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+import Models
+
+public enum RunningTimeEntryAction: Equatable {
+    case stopTapped(Int64)
```
Odrakir, line 5: This is part of another issue, I'm not sure if you need this in order to have at least a case in the enum. But in any case, you don't need to send the Int64, there's only one running TE. Look at the [module definition](https://www.notion.so/toggl/Running-Time-Entry-21542e11c3184711a50dc3e8bf031781) in notion.

Modules/Timer/Timer/RunningTimrEntry/RunningTimeEntryReducer.swift
```diff
@@ -0,0 +1,15 @@
...
+import Models
+import RxSwift
+import Repository
+import OtherServices
+
+func createRunningTimeEntryReducer(repository: TimeLogRepository, time: Time) -> Reducer<RunningTimeEntryState, RunningTimeEntryAction> {
+    return Reducer { _, action in
+        switch action {
+        default:
+            return []
```
Odrakir, line 12: Don't include defaults, we want the compiler to warn us when we add a new action

Modules/Timer/Timer/RunningTimrEntry/ViewModel/RunningTimeEntryViewModel.swift
```diff
@@ -0,0 +1,26 @@
+import UIKit
+import Models
+import Utils
+
+public struct RunninTimeEntryViewModel {
```
Odrakir, line 5: Why do we need a specific stuct for this? Why not using `TimeEntryViewModel?

Modules/Timer/Timer/RunningTimrEntry/RunningTimeEntrySelector.swift
```diff
@@ -0,0 +1,31 @@
...
+
+    return RunninTimeEntryViewModel(
+        timeEntry: timeEntry,
+        project: state.entities.getProject(timeEntry.projectId),
+        client: state.entities.getClient(project?.clientId),
+        task: state.entities.getTask(timeEntry.taskId)
+    )
+}
+
+let toElapsedTimeString: (RunninTimeEntryViewModel?, Date) -> String = { teVM, currentTime in
```
Odrakir, line 20: This shouldn't be here, it should be in an extension in the Utils module. But in any case... I don't think it's needed because there's an extension to `TimerInterval` already there.

Modules/Timer/Timer/RunningTimrEntry/Views/MarqueeView.swift
```diff
@@ -0,0 +1,48 @@
+import UIKit
+import CoreGraphics
+
+class MarqueeView: UIScrollView {
```
Odrakir, line 4: 🤔 Are we sure we need this? I think this is not 100% decided in the designs, I feel this should be added only when we know we needed and reviewed then, there's already a bunch of stuff going on in this PR.

Modules/Timer/Timer/RunningTimrEntry/Views/RunningTimeEntryViewController.swift
```diff
@@ -0,0 +1,46 @@
...
+//            .map { _ in Date() }
+//            .publish()
+//
+//        Driver.combineLatest(
+//            runningTimeEntryViewModelObservable,
+//            currentTimeObservable,
+//            resultSelector: toElapsedTimeString
+//        )
+//            .drive(timerLabel.rx.text)
+//            .disposed(by: disposeBag)
```
Odrakir, line 44: 🔥  or do something with this

Modules/Utils/Utils/UI/UIViewController + Extensions.swift
```diff
@@ -19,4 +19,12 @@ public extension UIViewController {
...
         child.didMove(toParent: self)
     }
+
+    func uninstall(_ child: UIViewController) {
+        child.willMove(toParent: nil)
+
+        child.view.removeFromSuperview()
+
+        child.removeFromParent()
+    }
```
Odrakir, line 29: We don't need this for now

Modules/Assets/Assets/Storyboards/Timer.storyboard
sofokli1: Don't know, will check

Modules/Timer/Timer/Common/SimpleBottomSheet.swift
sofokli1: what is TF?

Modules/Timer/Timer/Common/Views/TimerViewController.swift
sofokli1: if we install both `editView` and `running TE view` those show on top of each other. Should I play with the `hidden` property in this case?

Modules/Timer/Timer/RunningTimrEntry/RunningTimeEntryAction.swift
sofokli1: I do not need a case, will 🔥

Modules/Timer/Timer/Common/SimpleBottomSheet.swift
Odrakir: Text field

Modules/Timer/Timer/Common/Views/TimerViewController.swift
Odrakir: The start/edit view will appear on top of the running TE view. Let's see how that looks, we can consider using hidden or animating the running TE view down later on.

Modules/Timer/Timer/RunningTimrEntry/ViewModel/RunningTimeEntryViewModel.swift
sofokli1: it is part of the log feature and I am not sure what should be the policy regarding depending on an other features models. I saw it at first but decided to have the feature stand on its own without depending on the log models. What do you think?

Modules/Assets/Assets/Storyboards/Timer.storyboard
sofokli1: turns out this may have been from copy pasting IB elements from old app xib. it is 🔥 now

Modules/Timer/Timer/RunningTimrEntry/Views/MarqueeView.swift
sofokli1: 🔥

Modules/Timer/Timer/RunningTimrEntry/ViewModel/RunningTimeEntryViewModel.swift
Odrakir: I understand the reasoning, but that implies we will end up having a lot of duplicated code (mostly formatting code).

Modules/Assets/Assets/Storyboards/Timer.storyboard
Odrakir: Please, try not to do that when ever possible. I'm not saying we have to completely start from scratch, but this is a good moment to do some cleaning and have better lighter code an UI.

Modules/Timer/Timer/Common/SimpleBottomSheet.swift
```diff
@@ -24,36 +20,22 @@ class SimpleBottomSheet: UIViewController {
...
+        
         guard let parent = parent else { return }
 
         view.leadingAnchor.constraint(equalTo: parent.view.leadingAnchor).isActive = true
         view.trailingAnchor.constraint(equalTo: parent.view.trailingAnchor).isActive = true
-        bottomConstraint = view.bottomAnchor.constraint(equalTo: parent.view.safeAreaLayoutGuide.bottomAnchor)
-        heightConstraint = view.heightAnchor.constraint(equalToConstant: stops.first!)
-        bottomConstraint.isActive = true
-        heightConstraint.isActive = true
+        view.bottomAnchor.constraint(equalTo: parent.view.bottomAnchor).isActive = true
```
Odrakir, line 35: But now this view doesn't have a top anchor at all

Modules/Timer/Timer/Common/Views/TimerViewController.swift
```diff
@@ -28,26 +28,15 @@ public class TimerViewController: UIViewController {
...
-        store.select(isRunningTimeEntry)
-            .drive(onNext: updateRunningTimeEntryButtomSheet)
-            .disposed(by: disposeBag)
+        bottomSheet = BottomSheet(viewController: startEditViewController)
+        install(bottomSheet, customConstraints: true)
     }
     
     public override func viewDidLayoutSubviews() {
         super.viewDidLayoutSubviews()
         timeLogViewController.additionalSafeAreaInsets.bottom = bottomSheet.view.frame.height
```
Odrakir: At the same time, this should get the insets from the new view:
`timeLogViewController.additionalSafeAreaInsets.bottom = runningTimeEntryBottomSheet.view.frame.height`

Modules/Timer/Timer/RunningTimrEntry/Views/RunningTimeEntryViewController.swift
```diff
@@ -22,25 +22,5 @@ public class RunningTimeEntryViewController: UIViewController, Storyboarded {
 
     public override func viewDidLoad() {
         super.viewDidLoad()
```
Odrakir, line 24: This VC should have at least the preliminary UI

Modules/Assets/Assets/Storyboards/Timer.storyboard
sofokli1: The description label and the projectTaskClient label and the scroll view and the fade view looked a bit tricky and time consuming to do from scratch. Not my first choice to copy it for sure.

Modules/Timer/Timer/Common/SimpleBottomSheet.swift
sofokli1: it kind of never had, but I get the point of not having the vertical constrains setup. what do you suggest?

Modules/Timer/Timer/Common/SimpleBottomSheet.swift
```diff
@@ -0,0 +1,50 @@
...
+        super.viewDidLayoutSubviews()
+
+        guard !layedOut else { return }
+        layedOut = true
+
+        view.layer.cornerRadius = 10
+        view.layer.shadowColor = UIColor.black.withAlphaComponent(0.4).cgColor
+        view.layer.shadowOffset = CGSize(width: 0, height: -6)
+        view.layer.shadowRadius = 10.0
+        view.layer.shadowOpacity = 0.2
```
arturs105, line 48: Why not just put this whole thing (sans the `layedOut` part) inside `viewDidLoad`?

Modules/Timer/Timer/Common/SimpleBottomSheet.swift
```diff
@@ -0,0 +1,50 @@
...
+import Assets
+
+class SimpleBottomSheet: UIViewController {
+
+    private var layedOut: Bool = false
+
+    private let containedViewController: UIViewController
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
```
arturs105, line 12: Why have an unimplemented constructor in the first place?

sofokli1: if you do not put it there, xcode will hit you with a `'required' initializer 'init(coder:)' must be provided by subclass of 'UIViewController'`

Modules/Timer/Timer/RunningTimrEntry/Views/RunningTimeEntryViewController.swift
```diff
@@ -0,0 +1,58 @@
...
+
+        let runningTimeEntryViewModelObservable = store.select(runningTimeEntryViewModelSelector)
+
+        runningTimeEntryViewModelObservable
+            .map { $0?.description }
+            .drive(descriptionLabel.rx.text)
+            .disposed(by: disposeBag)
+
+        Observable<Int>
+            .timer(0, period: 1, scheduler: MainScheduler.instance)
```
Odrakir, line 36:
```suggestion
            .timer(.seconds(0), period: .seconds(1), scheduler: MainScheduler.instance)
```

Modules/Timer/Timer/RunningTimrEntry/Views/RunningTimeEntryViewController.swift
```diff
@@ -0,0 +1,58 @@
...
+            .asDriver(onErrorJustReturn: "00:00:00")
+            .drive(timerLabel.rx.text)
+            .disposed(by: disposeBag)
+
+        store.select(isRunningTimeEntrySelector)
+            .drive(onNext: updateUI)
+            .disposed(by: disposeBag)
+    }
+
+    private func updateUI(isRunningTimeEntry: Bool) {
```
Odrakir, line 54: Shouldn't we be starting/stopping the timer here? Maybe it's worth it to just use a regular Swift Timer on the current run loop instead of the RxSwift operator

PR #91, opened 2020-04-07
Introduce the CommitDeletion action

### LINKED ISSUE #64: Introduce the CommitDeletion action
## Description
As discussed [here](https://github.com/toggl/mobile-discussions/issues/58) we need an action that "flushes" the pending changes to the repo. We need such action because we want to check that we are deleting the correct set of time entries.

*NOTE: We are adding this action before it's actually emitted by any effect. This is needed to keep issues as small as possible and doesn't affect the testability of how the reducer will interact with the action, so this is fine*

## Dependencies
Depends on #63 

## Definition of done
- [x] The `CommitDeletion` action is added
- [x] The action is handled by the `TimeEntriesLogReducer` as shown in the linked mobile-discussions post
  - [x] A check is performed to ensure that the entries about to be committed are still the ones in the list of entries pending deletion
  - [x] If they are, empty the list of entries pending deletion and proceed to delete
  - [x] If they aren't, do nothing
- [x] Tests are added to ensure the reducer behavior is correct

sofokli1:
Closes #64

Not much code int reducer. Mostly check the tests. In the reducer I am sorting the array of TEs to delete because test fail randomly since the `Set` contained the IDs can be mapped in random order for obvious reasons, thus making it hard to test.


Modules/Timer/Timer/TimeEntriesLog/TimeEntriesLogReducer.swift
```diff
@@ -58,6 +59,18 @@ func createTimeEntriesLogReducer(repository: TimeLogRepository, time: Time) -> R
             
         case let .setError(error):
             fatalError(error.description)
+
+        case let .commitDeletion(timeEntryIds):
+            let timeEntryIdsToDelete = state.entriesPendingDeletion == timeEntryIds
```
heytherewill, line 64: Does equality comparison between HashSets ensure that the contents are the same?

Modules/Timer/Timer/TimeEntriesLog/TimeEntriesLogReducer.swift
```diff
@@ -58,6 +59,18 @@ func createTimeEntriesLogReducer(repository: TimeLogRepository, time: Time) -> R
...
+                ? timeEntryIds
+                : []
+
+            guard !timeEntryIdsToDelete.isEmpty else { return [] }
+
+            state.entriesPendingDeletion.removeAll()
+            return timeEntryIdsToDelete.sorted().map {
```
heytherewill, line 71: Why sort?

sofokli1: since the` timeEntryIdsToDelete` is a `set` there is no order. this makes the test expecting the delete call work some time and fails some time based on the order of ids. check `testCommitDeletionDeletesMultipleEntries`

Modules/Timer/Timer/TimeEntriesLog/TimeEntriesLogReducer.swift
sofokli1: apple says: `Use the “equal to” operator (==) to test whether two sets contain the same elements.` https://developer.apple.com/documentation/swift/set

Modules/Timer/Timer/TimeEntriesLog/TimeEntriesLogReducer.swift
sofokli1: if you have a better idea, I am all ears

sofokli1:
issues created #93

PR #94, opened 2020-04-08
Introduce the WaitForUndoEffect

### LINKED ISSUE #65: Introduce the WaitForUndoEffect
## Description
We need to add the effect that will actually delete time entries that are pending because of the undo.

## Dependencies
This PR needs the `CommitDeletion` action, introduced on #64 

## Definition of done
- [x] The effect that asynchronously waits for [5 seconds](https://github.com/toggl/mobileapp/blob/develop/Toggl.Core/Helper/Constants.cs#L26) and then emit a `CommitDeletion` action is added
- [x] Make actions that delete (swipes) emit the following:
  - [x] If there are no entries in the `entriesPendingDeletion` property, simply emit the `WaitForUndo` effect rather than emitting delete effects immediatelly
  - [x] If there are entries in the  `entriesPendingDeletion` property, emit effects to delete each of those entries and also the `WaitForUndo` effect.
- [x] Tests that cover all this new behavior are added

sofokli1:
Closes #65


Modules/Timer/Timer/TimeEntriesLog/TimeEntriesLogReducer.swift
```diff
@@ -19,7 +19,18 @@ func createTimeEntriesLogReducer(repository: TimeLogRepository, time: Time) -> R
...
+                    state.entriesPendingDeletion = [timeEntryId]
+                    return [waitForUndoEffect([timeEntryId])]
+                } else {
+                    let teIdsToDeleteImmediately = state.entriesPendingDeletion
+                    state.entriesPendingDeletion = [timeEntryId]
+                    var actions = teIdsToDeleteImmediately.sorted().map {
+                        deleteTimeEntry(repository, timeEntryId: $0)
+                    }
+                    actions.append(waitForUndoEffect([timeEntryId]))
+                    return actions
```
heytherewill, line 32: I think it'd be nice to extract this into a method so you don't need to copy the code twice.

Modules/Timer/Timer/TimeEntriesLog/TimeEntriesLogReducer.swift
```diff
@@ -94,3 +115,9 @@ private func continueTimeEntry(_ repository: TimeLogRepository, time: Time, time
             catch: { TimeEntriesLogAction.setError($0.toErrorType())}
     )
 }
+
+private func waitForUndoEffect(_ entriesToDelete: Set<Int64>) -> Effect<TimeEntriesLogAction> {
+    return Observable.just(TimeEntriesLogAction.commitDeletion(entriesToDelete))
+        .delay(RxTimeInterval.seconds(5), scheduler: MainScheduler.instance)
```
heytherewill, line 121: Extract that 5 into a constant that can be reused in the UI

heytherewill: That scheduler needs to be injected otherwise you won't be able to test it

Modules/Timer/Timer/TimeEntriesLog/TimeEntriesLogReducer.swift
sofokli1: an other action?

Modules/Timer/Timer/TimeEntriesLog/TimeEntriesLogReducer.swift
sofokli1: Was thinking of making some issues for the scheduler and how to test with scheduler to be done separately. what do you think?

Modules/Timer/Timer/TimeEntriesLog/TimeEntriesLogReducer.swift
heytherewill: No, not another action, just a method so that you don't need to copy the same code here and on the swipe group action

Modules/Timer/Timer/TimeEntriesLog/TimeEntriesLogReducer.swift
heytherewill: I'm fine with that. Create two issues: One for swapping this scheduler with an injected one (s) and one for creating the test infra for testing with schedulers (m)

Modules/Timer/Timer/TimeEntriesLog/TimeEntriesLogReducer.swift
```diff
@@ -100,6 +77,26 @@ func createTimeEntriesLogReducer(repository: TimeLogRepository, time: Time) -> R
     }
 }
 
+private func swipeLeftDelete(
```
heytherewill, line 80: The name is a bit misleading because we will eventually have features that change the direction. `deleteWithUndo` or something like that might be better

PR #99, opened 2020-04-08
🚧 Show a snackbar to allow undoing

### LINKED ISSUE #90: Show a snackbar to allow undoing time entry deletions
## Description
We need to show a snackbar to allow the user to undo deletions

## Dependencies
Requires #66 

## Definition of done
- [x] We display a Snackbar for 5 seconds whenever the list of pending deletions changes
- [x] The Snackbar should show the same string from the Xamarin app + an undo button
- [x] The undo button emits an `UndoButtonPressed` action
- [x] Pressing the button makes the deleted entries reappear on the log immediatelly
- [x] Pressing the button cancels the deletion of all entries pending deletion

sofokli1:
Closes #90


Modules/Timer/Timer/TimeEntriesLog/Views/TimeEntriesLogViewController.swift
```diff
@@ -74,8 +75,24 @@ public class TimeEntriesLogViewController: UIViewController, Storyboarded {
...
+                    self?.snackBar(for: setOfIds)
+                }
+                .drive(onNext: { $0?.show(in: self, duration: TimeInterval(TimerConstants.timeEntryDeletionDelaySeconds)) })
+                .disposed(by: disposeBag)
         }
     }
+
+    private func snackBar(for setOfIds: Set<Int64>) -> Snackbar {
+        return Snackbar.with(
+            text: setOfIds.count > 1 ? "\(setOfIds.count) time entries deleted" : "Time entry was deleted",
```
heytherewill, line 91: Extract these strings into resources for making i18n easier

Modules/Timer/Timer/TimeEntriesLog/Views/TimeEntriesLogViewController.swift
```diff
@@ -74,8 +75,24 @@ public class TimeEntriesLogViewController: UIViewController, Storyboarded {
...
+            buttonTitle: "UNDO",
```
heytherewill, line 92: Same thing regarding resources

Modules/Timer/Timer/TimeEntriesLog/Views/TimeEntriesLogViewController.swift
```diff
@@ -74,8 +75,26 @@ public class TimeEntriesLogViewController: UIViewController, Storyboarded {
...
             tableView.rx.setDelegate(self)
                 .disposed(by: disposeBag)
+
+            store.select(entriesPendingDeletionSelector)
+                .filter { !$0.isEmpty }
+                .map { [weak self] setOfIds -> Snackbar? in
+                    self?.snackBar(for: setOfIds)
+                }
+                .drive(onNext: { $0?.show(in: self, duration: TimeInterval(TimerConstants.timeEntryDeletionDelaySeconds)) })
+                .disposed(by: disposeBag)
```
Odrakir, line 85: I'm not convinced about this. We are adding too much logic to the view controller. And we are even including the timer, which is already in the reducer. Shouldn't we, instead, use the logic in the reducer and just show/hide the snackbar depending if `entriesPendingDeletionSelector` returns an empty set or not?

PR #104, opened 2020-04-13
🚧 Add suggestion models

### LINKED ISSUE #100: Create the Autocomplete suggestion type
## Description
We need a discriminated union to represent all possible autocomplete suggestions that can appear for the user. Since the view is new, we won't copy the [Xamarin implementation](https://github.com/toggl/mobileapp/tree/develop/Toggl.Core/Autocomplete/Suggestions) 1:1. The possible cases are:

- Time Entry Suggestion
- Project Suggestion
- Task Suggestion
- Tag Suggestion
- Create Project Suggestion
- Create Tag Suggestion

## Definition of done
- [ ] The type is added to our codebase

sofokli1:
Closes #100 

not sure about the last 2 points in the issue (`Create Project Suggestion`, `Create Tag Suggestion`) so didn't do anything for those.


heytherewill:
I'm not sure we need the protocol that's basically a marker interface. We need a proper discriminated union (in Swift's case, an enum) so that we are sure to always handle all possible cases

heytherewill:
Regarding the entity creation ones, the first option in the list of suggestions is `create project XXX`, so we need to consider that as well

Modules/Models/Models/Types/AutocompleteSuggestionType.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+public enum AutocompleteSuggestionType {
+    case timeEntrySuggestion
```
heytherewill, line 4: Each case should also hold the metadata needed for that entry to be created. For a timeEntrySuggestion, it needs to hold a time entry. For tasks, a project and a task. For the createProject, a string with the project name

heytherewill:
Droid implementation for reference https://github.com/toggl/android/pull/167/files

Modules/Models/Models/Types/AutocompleteSuggestionType.swift
```diff
@@ -1,10 +1,10 @@
...
 public enum AutocompleteSuggestionType {
-    case timeEntrySuggestion
-    case projectSuggestion
-    case taskSuggestion
-    case tagSuggestion
-    case createProjectSuggestion
-    case createTagSuggestion
+    case timeEntrySuggestion(timeEntry: TimeEntry)
+    case projectSuggestion(project: Project)
+    case taskSuggestion(task: Task)
```
heytherewill, line 6: The task suggestion also need the project data

sofokli1: It is hard to think what this cases need without actually having to use these

PR #107, opened 2020-04-13
Add scheduler provider service

### LINKED ISSUE #97: Inject scheduler in reducers
## Description
Inject the scheduler to reducers that need it so that those can be easily tested. This Scheduler may not be available yet, in that case create it.

## Dependencies
None

## Definition of done
- [x] Add the scheduler service (if not there already)
- [x] Inject it in the reducers that need it (TimeEntriesLogReducer maybe more)
- [x] Use the injected scheduler in those reducers

Odrakir:
### Summary
Adds the scheduler provider service, uses it were necessary and modifies tests.

### Technical considerations
The only complicated part around this is how I modified the `assertReducerFlow` so it takes all scheduled actions without blocking when the scheduler is using a delay. Any other ideas around that are welcome.

### Relationships
Closes #97 

### Pull Request checklist

### Pointers
Tap the description field and see the start/edit view open.

### Review template
- [ ] The code considers i18n

## Merge permissions
Go for it!


sofokli1:
this looks ok, is there a reason for keeping it as draft?

PR #109, opened 2020-04-14
Show start/edit view on state change

### LINKED ISSUE #89: Show start/edit view on state change
## Description
The start/edit view needs to appear when some events happen. This should be handled by reducers:
- When tapping the textfield in the running time entry view we show this view in the partial state (and editableTimeEntry should be nil)
- When tapping an entry in the log or the running time entry we show this view in the full state (and editableTimeEntry should include that TE details)
- When swiping down or hitting the close button we hide this view (this is missing an action for swipe down)

There's two ways we can do this:
- Adding a new route for the start/edit view
- Listening to change to the `EditableTimeEntry?` property in TimerState.

I think even if `EditableTimeEntry` has to be set in order for this view to do something at all, it's more semantically correct to use the route for this. So the reducer that wants to show this has to set both the route and the `EditableTimeEntry` property.

This might need the store also in the bottom sheet that holds the start/edit view, some refactoring might be needed.

## Dependencies
#38 

## Definition of done
- [ ] The new route is added
- [ ] Refactoring to listen to route changes and show/hide the view
- [ ] Change the route and the editable time entry property in reducers when starting a new entry and when editing an existing one
- [ ] Tests are added

Odrakir:
## Info
This PR makes the necessary changes so the start/edit view is toggled in and out of the screen when the user taps on a time entry (in the log or in the running time entry card) or taps the description in the start card. It also adds tests.

### Summary
Step by step this PR:
- Changes the coordination/viewcontroller hierarchy of the bottom sheet for start/edit
- Adds the `EditableTimeEntry` state to the log
- Adds and handles the `TimeEntryGroupTapped` action to the log
- Handles also group swipes (which was previously missing)
- Observes `editableTimeEntry` to show/hide the bottom sheet
- Binds buttons and gestures to dismiss the bottom sheet
- Displays the description of the TE being edited (if any)
- Adds tests

### Relationships
Closes #89 

### Pull Request checklist

### Pointers
Tap an entry in the log, it should open the start/edit view with that entry description. Tap the running entry, same thing. Tap the empty running entry (when there's no entry running), it should open the start/edit view without description.

Swipe the view down or hit the close button (top left), it should close that view.

Every commit is as small as possible, with the only problem that the last one (the merge) is a bit of a mess because we were touching the same files.

### Review template

- [x] The code is thoroughly tested in well written and resilient manner
- [x] The code is performant and leads to an optimal user experience
- [x] The code is testable and does not cut corners for the sake of delivering

- [x] There are no hard-coded numbers referring to dimensions or font sizes
- [x] The code considers UI elements that are similar and extracts their similarities into styles to avoid duplication

## Merge permissions
Do it


Odrakir:
Yeah, the handling of the focus was completely wrong. Please check again.

Modules/Timer/TimerTests/Reducers/StartEditReducerTests.swift
```diff
@@ -0,0 +1,9 @@
+//
+//  StartEditReducerTests.swift
+//  TimerTests
+//
+//  Created by Ricardo Sánchez Sotres on 15/04/2020.
+//  Copyright © 2020 Ricardo Sánchez Sotres. All rights reserved.
+//
+
+import Foundation
```
sofokli1: we have custom templates for this. and soon custom template for tests

PR #115, opened 2020-04-15
Setup telemetry services

### LINKED ISSUE #7: Setup telemetry services
## Description
We need to setup our telemetry services to emit events. This should be a reducer of its own.
We must use both AppCenter and Firebase. The approach is simple: we should have some injectable analytics service with a `track(event: Event)` method. This service will be a composite of two implementations (one for AppCenter and one for Firebase). Every app action must implement an `asEvent` extension method. Then we will have an `AnalyticsReducer` that takes the `analyticsService` as a dependency and handles all app actions to send events over the internet 🛰 

## Definition of done
- [x] The firebase and crashlytics telemetry packages are added to the project
- [x] All existing actions implement an `asEvent` extension method
- [ ] **All implementations of `asEvent` make sure the events are anonymous and that no user info is being sent to neither telemetry service**
- [x] The `AnalyticsReducer` is setup and working

dshr:
### Summary
This PR adds Firebase and Appcenter to the App and creates the Analytics reducer that helps us log all actions.

### Relationships
Closes #7

### Pull Request checklist

### Pointers
I mainly followed https://github.com/toggl/android/pull/42

### Review template

## Merge permissions
All yours!


Modules/TogglTrack/TogglTrack/App/AppAction.swift
```diff
@@ -80,3 +81,9 @@ extension AppAction: CustomDebugStringConvertible {
         }
     }
 }
+
+extension AppAction: Analytics.EventConvertible {
+    public func toEvent() -> Event {
+        return Event(self.debugDescription)
```
heytherewill, line 87: Let's not do that for now. We need to check @juanlaube's table of events in order to really get which event should be kept and which should not and we also need to ensure event names are precisely the same from the ones in the current app, so using the debugDescription here is dangerous

heytherewill: Also, not all actions emit events, so the `toEvent()` return type should be `Event?`

dshr: So just replace it with Event("Action event") for now?

Modules/OtherServices/OtherServices.xcodeproj/project.pbxproj
```diff
@@ -274,7 +274,7 @@
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
 				IPHONEOS_DEPLOYMENT_TARGET = 11.0;
-				MACH_O_TYPE = mh_dylib;
+				MACH_O_TYPE = staticlib;
```
Odrakir, line 277: Tests are failing because of this change. Why did you change this?

dshr: Thought it was a typo that I left, since all our frameworks are supposed to be static. I’ll change this asap

PR #127, opened 2020-04-16
Add StopButtonTapped action

### LINKED ISSUE #70: Add StopButtonTapped action
## Description
Action to stop the currently running time entry from the running time entry bottom card.

## Dependencies
#69 

## Definition of done
- [x] The new action is added
- [x] The new action is handled by the reducer and the effect returned
- [x] Tests are added

sofokli1:
### Summary
Adds the StopButtonTapped action and test. also connects the action to the UI

### Relationships
Closes #70

### Pull Request checklist

### Pointers
Just check the code and run it once to stop a TE

### Review template

## Merge permissions
🦑


Modules/Timer/Timer/RunningTimrEntry/RunningTimeEntryReducer.swift
```diff
@@ -6,7 +6,19 @@ import Repository
...
-    return Reducer { _, _ in
-        return []
+    return Reducer { state, action in
+        switch action {
+        case .stopButtonTapped:
+            guard var runningTimeEntry = state.runningTimeEntry else { return [] }
+            runningTimeEntry.duration = time.now().timeIntervalSince(runningTimeEntry.start)
+            state.entities.timeEntries[runningTimeEntry.id] = runningTimeEntry
+
+            stopTimeEntryEffect(state.entities.timeEntries[runningTimeEntry.id]!, repository: repository)
```
Odrakir, line 16: This should be returned, the effect. I know it's empty right now, but in this case just return `Effect.empty`

Also... are we setting the duration here or in the repository... 🤔 @heytherewill ?

sofokli1: I had it as `Effect.empty` but the assertion in the test was complaining

Modules/Timer/TimerTests/Reducers/RunningTimeEntryReducerTests.swift
```diff
@@ -0,0 +1,54 @@
...
+        let state = RunningTimeEntryState(entities: entities)
+
+        assertReducerFlow(
+            initialState: state,
+            reducer: reducer,
+            steps:
+            Step(.send, .stopButtonTapped) {
+                $0.entities.timeEntries[1]!.duration = 200
+            })
+    }
```
sofokli1, line 37: @Odrakir If I use an `Effect.empty` in stopTimeEntryEffect, the above test I get hit with this failure https://github.com/toggl/ios/blob/eugene/stopRunningTE/Modules/Timer/TimerTests/TestUtils/ReducerAssert.swift#L83

Modules/Timer/Timer/RunningTimrEntry/RunningTimeEntryReducer.swift
heytherewill: @Odrakir We're setting it in the repo, but I guess we should, instead, provide the stop date to the repo

Modules/Timer/Timer/Running/Views/RunningTimeEntryViewController.swift
```diff
@@ -35,6 +35,11 @@ public class RunningTimeEntryViewController: UIViewController, Storyboarded {
             .drive(onNext: updateUI)
             .disposed(by: disposeBag)
 
+        stopButtom.rx.tap
```
juanlaube, line 38: Just a small typo

```suggestion
        stopButton.rx.tap
```

PR #132, opened 2020-04-16
Add Project boilerplate

### LINKED ISSUE #110: Add the project module boilerplate
## Description
We need to add the boilerplate needed for the module responsible for adding new projects. We just need the boilerplate, no need to have concrete actions or handling.

One thing that needs to be added to the state is a `editableProject` type, similar to the editable time entry. That property is the one that will control the visibility of the project view

## Definition of done
- [x] The ProjectReducer is added
- [x] The ProjectState is added
- [x] The ProjectViewController is added
- [x] The coordinator boilerplate needed is added

heytherewill:
### Summary
This adds the basic boilerplate needed for work on the project module to start

### Technical considerations
I've added a little enum to avoid the string comparison that was happening in that view. I don't fully get why there's a list and all those casts going on when there could simply statically typed properties; I'll wait for @Odrakir to chime in and if there's no reason I'll just change that as well.

### Relationships
Closes #110 

### Pull Request checklist

### Pointers
Simple stuff and boilerplate, nothing too exciting to see here

### Review template

## Merge permissions
Anyone, anytime


Odrakir:
The enum is fine. I don't really like the name of the feature: "Project"... Project what? Selection? Creation?

heytherewill:
@Odrakir I does both. Since `StartEdit` is taken by the most prominent feature of the app (time entries), naming this `StartEditProject` is just needlessly verbose. There's no feature for selecting project in the new app, since the start/edit view will be unified.

Odrakir:
`ProjectManagment` then?

Modules/Timer/Timer/Common/TimerCoordinator.swift
```diff
@@ -8,15 +8,19 @@ public final class TimerCoordinator: BaseCoordinator {
...
-    public init(store: Store<TimerState, TimerAction>,
-                timeLogCoordinator: TimeEntriesLogCoordinator,
-                startEditCoordinator: StartEditCoordinator,
-                runningTimeEntryCoordinator: RunningTimeEntryCoordinator) {
+    public init(
+        store: Store<TimerState, TimerAction>,
+        timeLogCoordinator: TimeEntriesLogCoordinator,
+        startEditCoordinator: StartEditCoordinator,
+        runningTimeEntryCoordinator: RunningTimeEntryCoordinator,
+        projectCoordinator: ProjectCoordinator) {
```
sofokli1, line 18: a new line here would make this more readable

Modules/Timer/Timer/TimerFeature.swift
```diff
@@ -10,38 +10,54 @@ public func createTimerReducer(repository: Repository, time: Time) -> Reducer<Ti
...
-            timeLogCoordinator: (features["log"]!.mainCoordinator(store: store) as? TimeEntriesLogCoordinator)!,
-            startEditCoordinator: (features["startEdit"]!.mainCoordinator(store: store) as? StartEditCoordinator)!,
-            runningTimeEntryCoordinator: (features["runningTimeEntry"]!.mainCoordinator(store: store) as? RunningTimeEntryCoordinator)!
+            timeLogCoordinator: features[.timeEntriesLog]!.mainCoordinator(store: store) as! TimeEntriesLogCoordinator,
+            startEditCoordinator: features[.startEdit]!.mainCoordinator(store: store) as! StartEditCoordinator,
+            runningTimeEntryCoordinator: features[.runningTimeEntry]!.mainCoordinator(store: store) as! RunningTimeEntryCoordinator,
+            projectCoordinator: features[.project]!.mainCoordinator(store: store) as! ProjectCoordinator
         )
     }
+    // swiftlint:enable force_cast
```
sofokli1, line 60: shouldn't we keep all the linter comments at the top of the file, as we do in most files, just so we know where to look for these?

heytherewill:
@Odrakir This is a great name for the feature that lists projects and allows you to delete them in batch/see them all, like in webapp. I think Project is good enough as is for an add feature

Odrakir:
`AddProject` then? I'm saying this because `Project` is the model, the feature is supposed to do something with it, I think it could be confusing.

PR #136, opened 2020-04-20
Database refactor

### LINKED ISSUE #48: Add workpace persisting
## Description
We need to persist workspaces in disk properly

## Definition of done
- [ ] CoreData is properly configured to persist workspaces
- [ ] The repository is changed to use the actual workspaces
- [ ] The implementation, as discussed with @Odrakir, should follow the [principles of the Android implementation](https://github.com/toggl/android/blob/master/database/src/main/java/com/toggl/database/dao/WorkspaceDao.kt). This is to make this layer of app similar so that the sync team can act on it without much hurdle when context switching
- [ ] The Relationships between workspace and time entry are set
- [ ] The Relationships between workspace and project are set
- [ ] The load workspaces effect should should create a fake workspace if none is found on the DB. This is to simplify things until sync comes along. Everything, for now, should happen in a single workspace

### LINKED ISSUE #139: Add Task persisting
## Description
We need to persist tasks in disk properly

## Definition of done
- [ ] CoreData is properly configured to persist tasks
- [ ] Relationships are set

Odrakir:
## Info
This PR makes some fundamental changes to the DB layer. Basically we needed to use CoreData relationships instead of using IDs for relationships in the DB, but we still need to map to and from the normalized state we use in the rest of the app.

### Summary
The first thing we need to do is remove the repository/logic boundary for the models and instead have one that's between the DB and the repository. This way the repository will deal with structs and ID based relationships while the DB deals with NSManagedObject subclasses with CoreData relationships.

In order to make this tidy and reuse the most code I created different classes that split work from low level to higher level:
- `Managed`: All NSManagedObject subclasses conform to this protocol. It adds low leve core data operation methods to the objects. It also has the `prefetchRelationships` property which indicates the relationships we want to fetch when fetching the entity. We need all of its relationships (except inverse ones, which are not part of the subclass code) because we need the IDs of the related objects for mapping.
- `ManagedEntity`:  A specialisation of the previous that adds an ID an a method to fetch a specific NSManagedObject subclass by ID.
- `CoreDataModel`: Business logic models need to conform to this protocol and implement a couple of methods to map to and from NSManagedObject subclasses. It adds CRUD methods to those models.
- `CoreDataStack`: Just that, sets up core data. For now it exposes just the view context, but it's easy to add a background context for background operations
- `EntityDatabase`: It's generic over the entity type and adds all methods that the repository might need to interact with the DB. There's an extension specific to `TimeEntry`, we might need to create more extensions for other entities.

There are some things missing, let's see reviews for this and I can then create issues:
- Tests
- Handling errors (for now I'm jus throwing whatever, we could be more specific)
- Background DB operations... I think we will eventually need this

### Relationships
Closes #48 
Closes #139 

### Pull Request checklist

### Pointers
I think with the above explanation and how tidy the commit history is this should be "easy" to review. We probably don't need to do functional testing at this point (unless you have some data you can play with)

### Review template

- [ ] The code considers i18n

## Merge permissions
Me


Modules/Database/Database.xcdatamodeld/Database.xcdatamodel/contents
```diff
@@ -1,37 +1,66 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="16119" systemVersion="19E266" minimumToolsVersion="Automatic" sourceLanguage="Swift" userDefinedModelVersionIdentifier="">
+<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="16114.2" systemVersion="19D76" minimumToolsVersion="Automatic" sourceLanguage="Swift" userDefinedModelVersionIdentifier="">
     <entity name="ManagedClient" representedClassName="ManagedClient" syncable="YES">
         <attribute name="id" attributeType="Integer 64" defaultValueString="0" usesScalarValueType="YES"/>
         <attribute name="name" attributeType="String"/>
-        <attribute name="workspaceId" attributeType="Integer 64" defaultValueString="0" usesScalarValueType="YES"/>
+        <relationship name="inv_projects" toMany="YES" deletionRule="Nullify" destinationEntity="ManagedProject" inverseName="client" inverseEntity="ManagedProject"/>
```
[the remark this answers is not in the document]
dshr: Same thing, the inverse nature of the relationship is kind of implied, isn't it?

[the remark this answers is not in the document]
Odrakir: These are relationships that we don't actually need, but Apple recommends having them when using CoreData. If you look at the `NSManagedObject`subclasses, you'll see that I'm not including them there. I wanted to name them somehow to be clear this is not a mistake.

Any other ideas? The name has to start with a letter, so "_projects" is not an option.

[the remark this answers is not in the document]
dshr: Why can't we have just `projects`?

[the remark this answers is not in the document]
Odrakir: My reasoning is so it's not taken for a normal relationship that we DO need in the NSManagedObject subclass. Do you think it's not necessary to make that distinction?

Odrakir:
Implementing background execution is not hard at all. More of an issue of what API we would like to add, wether it'll be Rx or callbacks.

Modules/Database/Database.xcdatamodeld/Database.xcdatamodel/contents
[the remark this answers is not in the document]
juanlaube: If we don't need the relationship then ignoring it in the NSManagedObject subclass is enough, however if we want to use it in the future for any particular reason, then we will need a migration to remove the prefix.
I'd just keep the relationship name in the model as if we were using it (e.g: use `projects` instead of `_projects` or `inv_projects`) and ignore its existence in the class.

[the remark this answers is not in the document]
Odrakir: Ok, I'm not completely convinced, but since you all agree I'll change that.

Modules/Database/Database/TimeEntryDatabase.swift
```diff
@@ -0,0 +1,20 @@
...
+public class TimeEntryDatabase: EntityDataBase<TimeEntry> {
+
+    public func getAllSorted() throws -> [TimeEntry] {
+        try TimeEntry.get(
+            in: stack.viewContext,
+            sortDescriptors: [NSSortDescriptor(key: "start", ascending: false)]
+        )
+    }
+
+    public func getAllRunning() throws -> [TimeEntry] {
```
sofokli1: why use an array for this, are we allowing for multiple running TEs?

Modules/Database/Database/CoreData/Managed.swift
```diff
@@ -0,0 +1,62 @@
...
+
+        return object
+    }
+
+    static func fetch(
+        in context: NSManagedObjectContext,
+        configurationBlock: (NSFetchRequest<Self>) -> () = { _ in }
+    ) -> [Self] {
+        let request = NSFetchRequest<Self>(entityName: Self.entityName)
+        request.relationshipKeyPathsForPrefetching = prefetchRelationships
```
sofokli1, line 35: what if instead of `relationshipKeyPathsForPrefetching` we go for `propertiesToFetch`? This works also for to-one relationships. In a way we mostly use the relation to get the id. So we could add all the attributes of the entity and the `relation.id` this way it will still fetch the relation but only the id will be available the rest will be faults. I am ok with the current solution and do not think that it will affect performance. In case you want to do the best optimisation from the get go, this other way kind of makes more sense. I am not asking for changes just starting a conversation.

Modules/Database/Database/Models/ManagedObjects/ManagedTag.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+import CoreData
+
+public final class ManagedTag: NSManagedObject, ManagedEntity {
+    @NSManaged var id: Int64
+    @NSManaged var name: String
+
+    @NSManaged var workspace: ManagedWorkspace
```
sofokli1, line 8: if we only keep this for the id, why not make it private and have a computed var that returns the id?

Modules/Database/Database/TimeEntryDatabase.swift
Odrakir: I'm not sure if that piece of logic should be here, in the repo or in the reducers. For now I took the simpler route. We can change the API later on if we feel we need it.

Modules/Database/Database/CoreData/Managed.swift
Odrakir: 🤔 That wouldn't work for to many relationships like `tags` in `TimeEntry`, right?

Modules/Database/Database/Models/ManagedObjects/ManagedTag.swift
Odrakir: While that would make the code safer, I think it also makes it more complicated. Let's leave it for now.

Modules/Database/Database/CoreData/Managed.swift
sofokli1: forgot about the tags

PR #137, opened 2020-04-20
Show the start/edit view when the description field is tapped 

### LINKED ISSUE #133: Show the start/edit view when the description field is tapped
## Description
Show the start TE view then the description field on the running time entry bottom card is tapped (when there's no time entry running).

## Dependencies
#69 

## Definition of done
- [ ] Focusing the description field triggers the `.cardTapped` action.

dshr:
### Summary
This PR opens the start/edit view when the description field is tapped.

### Technical considerations
This is done by setting the `userInteractionEnabled` property to false for the TextField, so the tap events get caught by the parent view instead.

This seems to be the simplest way yet. My only concern would be accessibility support for this, but maybe for that case we could merge the running te card with the start/edit card, at least at the code level.

### Relationships
Closes #133

### Pull Request checklist

### Pointers
Tap the description field and see the start/edit view open.

### Review template
- [ ] The code considers i18n

## Merge permissions
Go for it!


dshr:
I'd advice against that, since this might work weird with screen reader, but it really depends on the final UI.

PR #140, opened 2020-04-21
📐 Add ProjectViewController layout

### LINKED ISSUE #111: Add the basic layout for the project view
## Description
We need to [add the basic layout](https://toggl.invisionapp.com/d/main#/console/19353169/404203834/preview) for the add project view. This is to make it simpler to add the actions and test them.

For now, let's not focus on the colors, since they are a bit harder. Let's just add the basicaly layout with the chips/pills and call it a day.

## Definition of done
- [ ] The basic layout is implemented

juanlaube:
## Info
This adds the layout for the AddProjectViewController.

### Summary
As mentioned, this adds only the layout for the AddProjectViewController, doesn't care about presenting it yet, but it's done in such way that could be used with a BottomSheet.
The scene is in a new storyboard called `Project` to avoid conflicts.

### Relationships
Closes #111 

### Pull Request checklist

### Pointers
Since the view controller is not presented yet, testing is only possible if you do add the necessary code to present it somewhere.
Here's a screenshot from my tests:
Simulator Screen Shot - iPhone 11 Pro - 2020-04-21 at 11 17 00

### Review template
- [ ] The code considers i18n

## Merge permissions
🦑


Modules/Timer/Timer/AddProject/Views/AddProjectViewController.swift
```diff
@@ -0,0 +1,22 @@
+import UIKit
+import Assets
+import Utils
+
+public final class AddProjectViewController: UIViewController, Storyboarded {
```
heytherewill, line 5:
```suggestion
public final class ProjectViewController: UIViewController, Storyboarded {
```

PR #165, opened 2020-04-23
Add wheel math extensions

### LINKED ISSUE #147: Port Math utils used in the wheel from the old apps
## Description
This should be a straight port from the current implementation.
Let's take a look on what's being used, port it and be done with it.

## Dependencies
None.

## Definition of done
Basically, we need:
- [ ] The helper methods used in all wheel components are ported:
  - [ ] ToAngleOnTheDial
  - [ ] ToAngle
  - [ ] ToPositiveAngle
  - [ ] AngleToTime
  - [ ] AngleBetween
  - [ ] DistanceSq
  - [ ] PointOnCircumference
  - [ ] IsBetween
  - [ ] PingPongClamp
  - [ ] ToRadians
- [ ] The constants used in the wheel are also ported:
  - [ ] FullCircle
  - [ ] MinutesInAnHour
  - [ ] SecondsInAMinute

heytherewill:
### Summary
This ports the math extensions we had in Xamarin (used by the wheel) to Swift

### Technical considerations
I tried making things more Swifty:tm:, but feel free to point out places where it can be further improved

### Relationships
Closes #147 

### Pull Request checklist

### Pointers
Check if the port is correct and make sure that things translate correctly into Swift. Also, if anything can be improved, even if slightly, let me know.

### Review template

## Merge permissions
Anyone, anytime


Modules/Utils/Utils/Extensions/WheelMathExtensions.swift
```diff
@@ -0,0 +1,95 @@
...
+
+    func distanceSq(_ point1: CGPoint, _ point2: CGPoint) -> CGFloat {
+        let deltaX = point1.x - point2.x
+        let deltaY = point1.y - point2.y
+        return deltaX * deltaX + deltaY * deltaY
+    }
+}
+
+extension CGFloat {
+    static var tau: CGFloat { 2 * .pi }
```
juanlaube, line 20: 😂 𝛕 master race!

Modules/Utils/Utils/Extensions/WheelMathExtensions.swift
```diff
@@ -0,0 +1,95 @@
+import Foundation
+import CoreGraphics
+import Darwin
+
+protocol WheelMath { }
```
juanlaube, line 5: Are we going to have a type conforming to this protocol? 🤔

Modules/Utils/Utils/Extensions/WheelMathExtensions.swift
```diff
@@ -0,0 +1,95 @@
...
+
+extension WheelMath {
+    func angleBetween(_ point1: CGPoint, _ point2: CGPoint) -> CGFloat {
```
juanlaube, line 8: I'm being nitpick now, but for 110% swiftyness 👌 
```suggestion
    func angle(Between point1: CGPoint, and point2: CGPoint) -> CGFloat {
```

Modules/Utils/Utils/Extensions/WheelMathExtensions.swift
```diff
@@ -0,0 +1,95 @@
...
+        return atan2(point1.y - point2.y, point1.x - point2.x)
+    }
+
+    func distanceSq(_ point1: CGPoint, _ point2: CGPoint) -> CGFloat {
```
juanlaube, line 12: Same here

Modules/Utils/Utils/Extensions/WheelMathExtensions.swift
heytherewill: Yes, in any place that needs to use the `angle` method

Odrakir:
I know this comes from an Android need, but the process is a bit weird: Adding extensions that we yet don't need, and we might not need at all depending on how much our implementation of the wheel differs from the one in the current app.

Modules/Utils/Utils/Extensions/WheelMathExtensions.swift
juanlaube: I find it really weird tbh, but perhaps I'm not seeing something? Why not make these two methods as extensions on `CGPoint` where self is the first point? Something like this perhaps:
```
extension CGPoint {
    func angle(with other: CGPoint) -> CGFloat { ... }
    func distanceSq(with other: CGPoint) -> CGFloat { ... }
```

PR #166, opened 2020-04-23
🚧 Add DoneButtonTapped action to project module

### LINKED ISSUE #114: Add the DoneButtonTapped action
## Description
We need to add the action that creates a project with the current information. Right now the mobile app runs validation before saving, so when handling this action we should check whether the project can be created. We should basically check if there are no other projects in the same workspace that have the same name and the same client, so:

```swift
projects.none { $0.name == name && $0.workspaceId == workspaceId && $0.clientId == clientId  }
```

If the project can't be created, we need to set an error.

## Dependencies
Depends on both #110 and #111  

## Definition of done
- [x] The DoneButtonTapped action is added to the list of Project actions
- [x] The action is handled by the reducers
- [x] Proper validation is added
- [x] The behavior is tested
- [x] The action is emited by the UI

juanlaube:
## Info
Adds the doneButtonTapped action.

### Summary
Adds the doneButtonTapped action, handles it in the reducer and emits the createProjectEffect after validation.

### Technical considerations
Everything follows the implementation on Android.

### Relationships
Closes #114 

### Pull Request checklist

### Pointers
The important changes are in the reducer and tests, everything else is boilerplate and some extension methods.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


Modules/Timer/Timer/Project/ProjectReducer.swift
```diff
@@ -11,7 +11,37 @@ func createProjectReducer(repository: TimeLogRepository) -> Reducer<ProjectState
         switch action {
         case .nameEntered(let name):
             state.editableProject?.name = name
+            state.editableProject?.error = .none
+            return []
+
+        case .doneButtonTapped:
+            guard let editableProject = state.editableProject else { return [] }
+            guard isValid(project: editableProject, existing: state.projects) else {
+                state.editableProject?.error = .projectAlreadyExists
```
Odrakir, line 20: I'm not completely sure about this, but in other reducers, setError is an action so it goes through the system and we can trace it.

juanlaube: That's the approach I was taking, but this is how it was implemented on Android so I went with that.

Modules/Timer/Timer/Project/EditableProject.swift
```diff
@@ -9,6 +9,14 @@ public struct EditableProject: Equatable {
     
     public var workspaceId: Int64
     public var clientId: Int64?
+
+    public var error: ProjectError = .none
+
+    public enum ProjectError {
+        case none
```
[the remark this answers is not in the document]
juanlaube: I don't have a strong opinion about this and both look good to me, but since this is probably what most swift devs expect, I will go with the option.

Modules/Models/Models/Project.swift
```diff
@@ -28,12 +28,12 @@ public struct Project: Codable, Entity, Equatable {
     public init(
         id: Int64,
         name: String,
-        isPrivate: Bool,
-        isActive: Bool,
-        color: String,
-        billable: Bool?,
+        isPrivate: Bool = true,
+        isActive: Bool = true,
+        color: String = "#ffffff",
```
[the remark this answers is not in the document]
heytherewill: It will all change when the color project comes around, so dealing with specific colors is not that relevant. Since this will create unreadable project names, though, maybe consider a better colors 😂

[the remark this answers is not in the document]
juanlaube: I moved the defaults to the DTO.

Modules/Timer/Timer/Project/EditableProject.swift
```diff
@@ -10,10 +10,9 @@ public struct EditableProject: Equatable {
     public var workspaceId: Int64
     public var clientId: Int64?
 
-    public var error: ProjectError = .none
+    public var error: ProjectError? = .none
 
     public enum ProjectError {
```
Odrakir, line 15: But now this doesn't need to be an enum anymore (at least for now). Also, it should inherit from `Error`.

Also... 🤔 I'm not sure about it being an embedded type. What's your thinking there? We should definitely work on error management for the whole app.

juanlaube: I deleted the error and replaced it with a bool for now. However I disagree on it not being an nested type and having it to implement Error, first because the ProjectError alone doesn't make sense, it's tied to the EditableProject and nothing else, second forcing it to implement Error was unnecessary since we were not throwing it, ProjectError in this case just represents the error message we are supposed to show later in the UI.

Odrakir: But displaying error messages in the UI will be easier if all come from the same property in the state, and for that they need a common type.

Modules/Timer/Timer/Project/ProjectReducer.swift
```diff
@@ -11,10 +11,40 @@ func createProjectReducer(repository: TimeLogRepository) -> Reducer<ProjectState
...
+            guard let editableProject = state.editableProject else { return [] }
+            guard isValid(project: editableProject, existing: state.projects) else {
+                state.editableProject?.hasError = true
+                return []
+            }
+            return [createProjectEffect(state, repository)]
+
+        case .projectCreated(let newProject):
+            state.projects[newProject.id] = newProject
+            state.editableProject = nil
```
heytherewill, line 27: Won't this be perceived like the app not being responsive? 

e.g.: The user taps done and the view only disappears a split second afterwards, which is after the DB persisted and responded? 🤔

heytherewill: IOW, shouldn't this line be in `doneButtonTapped`?

Modules/Timer/Timer/Project/ProjectReducer.swift
```diff
@@ -11,10 +11,40 @@ func createProjectReducer(repository: TimeLogRepository) -> Reducer<ProjectState
...
+    return repository.createProject(dto)
+        .map(ProjectAction.projectCreated)
+        .toEffect()
+}
+
+func isValid(project newProject: EditableProject, existing projects: [Int64: Project]) -> Bool {
+    return projects.values.none { project in
+        project.name == newProject.name &&
+            project.workspaceId == newProject.workspaceId &&
+            project.clientId == newProject.clientId
```
heytherewill, line 48: This indentation is funny

PR #169, opened 2020-04-24
Refactor analytics integration

Odrakir:
## Info
The way we were integrating the analytics frameworks was problematic (mostly due to Firebase), they want to be in the main app target, but we wanted them to be in the analytics framework.

What this PR does is keep our analytics framework but move its dependencies to the main app target and injecting them into the analytics service.

There might be another way to fix the whole mumbo jumbo of frameworks, Cocoapods and all that, if anyone knows, I'm all ears. Until then I think this should work.

### Relationships
Doesn't close any issue, but I need this for the configuration files issue.

### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
Go!


Modules/Analytics/Analytics/Event.swift
```diff
@@ -1,8 +1,8 @@
 import Foundation
 
 public struct Event {
-    var name: String
-    var parameters: [String: Any]?
+    public var name: String
+    public var parameters: [String: Any]?
```
juanlaube, line 5:
```suggestion
    public let name: String
    public let parameters: [String: Any]?
```

PR #171, opened 2020-04-24
Add DateTimePickMode enum 

### LINKED ISSUE #143: Create DateTimePickMode enum and add it as a property to handle editing time entry start/end times to the StartEdit state
## Description
We need a new property added to the StartEdit state to help handling time/date/barrel pickers.
- `datetimePickMode ` should be an enum (DateTimePickMode) with the following possible values:
  - None -> indicates that we are not any picking time/date
  - Start -> for the start time barrel
  - End -> for the end time barrel

## Dependencies
None.

## Definition of done
- [ ] The enum `DateTimePickMode` for the `datetimePickMode` is created
- [ ] The `datetimePickMode` property is added to the StartEdit state

dshr:
### Summary
This PR creates the aforementioned enum and adds it to the StartEditState. The approach here is a bit different from the spec, since instead of the `.none` case we can use an optional, thus keeping a lighter state while we're not editing durations.

### Relationships
Closes #143 

### Pull Request checklist

### Pointers
Commit by commit works best here.

### Review template

## Merge permissions
All yours!


daividssilverio:
@sofokli1 the idea was to use a `None` enum entry as well, not `null`.

dshr:
Fair enough, `.none` it is then

PR #173, opened 2020-04-27
Add build configuration files

### LINKED ISSUE #128: Add build configuration files
## Description
Adding new projects and targets to the app is a bit messy. There's many changes to the build configuration that need to be done and it's very error prone.

This can be solved by using build configuration files in Xcode: [xcconfig files](https://nshipster.com/xcconfig/)

## Definition of done
- [ ] We end up with all the needed configuration files for static frameworks, dynamic frameworks, tests and the main app. I'm not sure how that would look, if we need 4 different files or not. Some research needs to be done
- [ ] Said configuration allows to add new modules without much hassle

Odrakir:
## Info
This PR adds xcconfig files for every target in the app (main app, frameworks and tests). Together with some other base xcconfig files to share settings.

The idea is we stop making changes to build settings through the Xcode plist editor and instead use these files which are much easier to understand and less prone to merge conflicts.

### Technical considerations
Every target build configuration file inherits from a base file and also optionally (with "#include?") from a CocoaPods file. That's needed for CocoaPods to work and if it's missing it'll let us know when running "pod install"

Next thing we should do is creating a static framework project template that generates those config files for us too.

### Relationships
Closes #128 

### Pull Request checklist

### Pointers

Commits start very granular and easy to understand, and get bigger over time. Just follow the yellow brick road.

### Review template

- [ ] The code considers i18n

## Merge permissions
Let me merge it


Odrakir:
This is kind of a mess to review, and I found a simper way. Let me close this for now.

PR #176, opened 2020-04-29
Display duration on start/edit view

### LINKED ISSUE #121: Properly display the current duration
## Description
We need to ensure that the duration on the top right corner of the card is correct
## Definition of done
- [ ] If the duration is set, we just use it along with the existing format method used in the log
- [ ] If there's no duration, we calculate the duration based on the current date and then use the existing format method
- [ ] If there's no duration, we need to update the number every second

Odrakir:
## Info
Displays the duration in the start edit view:
- When the TE is new and thus not started yet, it displays:  "00:00"
- When the TE is an existing stopped one it displays its duration in format: "h:mm:ss"
- When the TE is running (no duration) it displays "h:mm:ss" taking the current time from the time service

### Technical considerations
I changed the `formattedDuration` extension on `TimeInterval` which made no sense as it was taking a `Date` parameter.

### Relationships
Closes #121 

### Pull Request checklist

### Pointers
Should be pretty simple to review

### Review template
- [ ] The code considers i18n

## Merge permissions
Do it!


Modules/Timer/Timer/StartEdit/Views/StartEditViewController.swift
```diff
@@ -128,6 +137,28 @@ public class StartEditViewController: UIViewController, Storyboarded, BottomShee
         }
 
         descriptionTextField.text = timeEntry.description
+        setDurationLabel(for: timeEntry)
+    }
+
+    private func setDurationLabel(for timeEntry: EditableTimeEntry) {
+
+        timer?.invalidate()
+        timer = nil
```
sofokli1, line 146: Should we maybe have this 2 lines in the view `didDisappear` as well?

Modules/Utils/Utils/Extensions/TimeInterval+Extensions.swift
```diff
@@ -1,16 +1,12 @@
 import Foundation
 
 extension TimeInterval {
-    public func formattedDuration(ending endingDate: Date) -> String {
+    public func formattedDuration() -> String {
 
-        let startingDate = endingDate.addingTimeInterval(-self)
-        let calendar = Calendar.current
+        let formatter = DateComponentsFormatter()
```
sofokli1, line 6: can we extract this outside of the func. doesn't make sense to create the same thing every time we call the func.

Modules/Timer/Timer/StartEdit/Views/StartEditViewController.swift
Odrakir: 🤔 I think I should just invalidate the timer on deinit, right? If I do it on didDisappear, when ever we send the app to background and come back the timer would be stopped

sofokli1: Even better

PR #177, opened 2020-04-29
Include the cursor position in DescriptionEntered

### LINKED ISSUE #120: Update the DescriptionEntered action to include the cursor position
## Description
Since we are implementing the suggestion mechanism, the current position of the cursor is pretty important.
It doesn't need to be stored in the state, since it'll just be later used by an effect to calculate the suggestions
Emitting this action when the cursor changes should be handled in another issue.
## Definition of done
- [x] The action is updated to match the docs
- [x] The place where the action is dispatched is changed to return the cursor position as well 

sofokli1:
### Summary
Simply gets the cursor position in the VC and sends it to the action

### Relationships
Closes #120

### Pull Request checklist

### Pointers
Check the code, is very minimal.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


Modules/Timer/Timer/StartEdit/Views/StartEditViewController.swift
```diff
@@ -115,6 +117,11 @@ public class StartEditViewController: UIViewController, Storyboarded, BottomShee
             .disposed(by: disposeBag)
     }
 
+    private func descriptionCursorPosition() -> Int? {
+        guard let selectedRange = descriptionTextField.selectedTextRange else { return nil}
```
heytherewill, line 121:
```suggestion
        guard let selectedRange = descriptionTextField.selectedTextRange else { return nil }
```

PR #179, opened 2020-04-29
Add first analytics events

### LINKED ISSUE #178: Add first round of events
## Description
After [going through all events in the old apps](https://www.notion.so/toggl/Old-App-Events-that-need-mapping-5ffa065959f1405a8b5fc7128fb23840) I managed to compile a list of events that can be implemented with the code we already have. We should do so now and then implement future events as they come.

## Definition of done
- [ ] The `EditViewClosed` event is implemented with the same parameters as the Xamarin app
- [ ] The `EditViewOpened` event is implemented with the same parameters as the Xamarin app
- [ ] The `TimeEntryStopped` event is implemented with the same parameters as the Xamarin app
- [ ] The `DeleteTimeEntry` event is implemented with the same parameters as the Xamarin app

heytherewill:
### Summary
This adds the first round of analytics events to get our codebase up to speed. Other events can later be added alongside with actions.

### Relationships
Closes #178

### Pull Request checklist
// I'm not a fan of testing Analytics events. It kinda becomes silly to the point where you copy the analytics code to check that a method was called. For that reason I'm not adding tests for this.

### Pointers
Check commit by commit to reducer clutter. The last commit is the important one.

### Review template
- [x] The code is thoroughly tested in well written and resilient manner
- [x] The code is performant and leads to an optimal user experience
- [x] The code is testable and does not cut corners for the sake of delivering
- [X] The code considers older platforms well and will work in older devices just like it will in new ones

## Merge permissions
Anyone, anytime


Modules/TogglTrack/TogglTrack/AnalyticsReducer.swift
```diff
@@ -1,14 +1,78 @@
...
+extension AppAction {
+    public func toEvent(_ state: AppState) -> Event? {
+        switch self {
+        case let .timer(.startEdit(startEditAction)):
+            return startEditAction.toEvent(state)
+        case let .timer(.runningTimeEntry(runningAction)):
+            return runningAction.toEvent()
+        case let .timer(.timeLog(logAction)):
+            return logAction.toEvent()
+        default:
```
Odrakir, line 27: The default case makes this error prone when adding new actions. I would rather return nil explicitly for cases that send no events.

Modules/Analytics/Analytics/Event.swift
```diff
@@ -4,12 +4,30 @@ public struct Event {
...
+    private init(_ name: String, parameters: [String: Any]? = nil) {
         self.name = name
         self.parameters = parameters
     }
-}
-
-public protocol EventConvertible {
-    func toEvent() -> Event?
+    
+    public static func editViewClosed(_ reason: EditViewCloseReason) -> Event {
```
Odrakir, line 12: While I see how these methods is useful I find it not ideal that we have to modify two files to add new events... 🤔 Can we move this to the action extensions?

Modules/TogglTrack/TogglTrack/AnalyticsReducer.swift
heytherewill: This is on purpose becausemost actions won't have an event. We don't need to edit this file to add a nil return for every action, since only a handful will contain events

Modules/Analytics/Analytics/Event.swift
heytherewill: I thought about this as well, and the problem this creates is that will scatter events throughout the entire system. Take the `editViewOpen` event, for example. It will be created in both the calendar and the timer modules. If we create those events in two completely separate files, we will need to rely on reviews to verify that the same event is using the same pair of name and parameters on every usage inside the app. If we force events to use the static methods, they will have consistent naming.

Modules/TogglTrack/TogglTrack/AnalyticsReducer.swift
Odrakir: Your call, but I think we'll end up missing events

Modules/TogglTrack/TogglTrack/AnalyticsReducer.swift
```diff
@@ -1,14 +1,78 @@
...
+            return nil
+        }
+    }
+}
+
+extension StartEditAction {
+    public func toEvent(_ state: AppState) -> Event? {
+        switch self {
+        case .closeButtonTapped, .dialogDismissed:
+            return Event.editViewClosed(.close)
```
Odrakir, line 37: I think you can remove the `Event` part to make it a bit more concise.

Modules/Analytics/Analytics/Event.swift
Odrakir: Ok then.

PR #180, opened 2020-04-29
Add DurationInputted action to the StartEdit reducer

### LINKED ISSUE #152: Add the DurationInputted action to the StartEdit reducer
## Description
This action is emitted when the user is inputting valid durations in the manual duration input view from The Wheel™.
- it should change the duration of the time entry being edited `editableTimeEntry` (start time if running, end time otherwise)
- invalid durations should be filtered out

## Definition of done
- [x] The action is added
- [x] If the time entry is running, changing the duration should change the start time of the `editableTimeEntry`
- [x] If the time entry is stopped, changing the duration should change the actual duration of the `editableTimeEntry`
- [x] Invalid durations are filtered out 
- [x] Tests are added
- [x] The [documentation](https://www.notion.so/toggl/DurationInputted-31331ad5d01a45bea600a2075bcd0a16) is updated to reflect the changes done in this PR

arturs105:
### Summary
Adds and handles `StartEditAction.DurationInputted`.

### Relationships
Closes #152

### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


Modules/Timer/Timer/StartEdit/StartEditReducer.swift
```diff
@@ -59,6 +60,17 @@ func createStartEditReducer(repository: TimeLogRepository, time: Time) -> Reduce
             state.editableTimeEntry?.description = appendCharacter("#", toString: editableTimeEntry.description)
             return []
 
+        case let .durationInputted(duration):
+            guard duration >= 0 else { return [] }
+            guard let editableTimeEntry = state.editableTimeEntry else { return [] }
+            if editableTimeEntry.isGroup { return [] }
```
Odrakir, line 66:
```suggestion
            guard duration >= 0,
                let editableTimeEntry = state.editableTimeEntry,
                !editableTimeEntry.isGroup
                else { return [] }
```

Modules/Timer/Timer/StartEdit/StartEditReducer.swift
```diff
@@ -5,6 +5,7 @@ import RxSwift
 import Repository
 import OtherServices
 
+// swiftlint:disable function_body_length
 // swiftlint:disable cyclomatic_complexity
```
Odrakir, line 9: These two can go in the same line
```suggestion
// swiftlint:disable cyclomatic_complexity function_body_length
```

Modules/Timer/Timer/Common/EditableTimeEntry.swift
```diff
@@ -63,4 +63,12 @@ extension EditableTimeEntry {
     func toStartTimeEntryDto() -> StartTimeEntryDto {
         return StartTimeEntryDto(workspaceId: self.workspaceId, description: self.description)
     }
+
+    var isRunning: Bool {
```
Odrakir, line 67: To be more precise, shouldn't this be called `isRunningOrNew`?

PR #182, opened 2020-04-30
Add the DateTimePicked action to the StartEdit reducer 

### LINKED ISSUE #146: Add the DateTimePicked action to the StartEdit reducer
## Description
This action takes in the value obtained from the picker.
This action should be emitted when a datetime is picked from one of the time/date/barrels has a value selected.
The reducer should validate the input from the action to prevent inconsistencies.
The reducer should set the `editableTimeEntry`'s start time/duration accordingly.

*On Android, this action should also set the `datetimePickMode` to `None` after executing successfully.
*On iOS, this action does do anything else, as the users can keep playing with the barrel until they are satisfied with it.

## Dependencies
Depends on #143  

## Definition of done
- [ ] The action is added and takes in a datetime parameter
- [ ] The `editableTimeEntry`s start time and duration is updated accordingly based on the current `datetimePickMode`
- [ ] On android, the action input needs to be validated to make sure the `start time` and `duration` are valid. An exception should be thrown if the input is invalid. Check the [old source code for reference]. This might not be needed on iOS. (https://github.com/toggl/mobileapp/blob/2fba1cdea3fa55f107f331d83972f3a677e5509b/Toggl.Core.UI/ViewModels/EditDurationViewModel.cs#L275-L307).
- [ ] The behavior is tested 
- [ ] The [documentation](https://www.notion.so/toggl/DateTimePicked-8c45d89481fc4fd99381771c46090065) is updated to reflect the changes done in this PR

dshr:
### Summary
This PR add the dateTimePicked action to the start edit reducer and sets the start time and duration of the editableTimeEntry? appropriately.

### Relationships
Closes #146

### Pull Request checklist

### Pointers
Commit by commit should work well enough

### Review template

## Merge permissions
All yours!


Modules/Timer/Timer/StartEdit/StartEditReducer.swift
```diff
@@ -90,6 +98,15 @@ func doneButtonTapped(_ state: StartEditState, _ repository: TimeLogRepository)
     return []
 }
 
+func dateTimePicked(_ state: inout StartEditState, date: Date) {
+    if state.editableTimeEntry == nil { return }
```
heytherewill, line 102: `guard let` and throw if we reach this invalid state

dshr: We're mutating the editable TE here, so a `guard let` doesn't really make much sense, I just added the throw here.

PR #199, opened 2020-05-08
Styling framework proposal

dshr:
My Q2 OKR is creating a reusable components system for our apps. This will allow us to abstract away a lot of the boilerplate we normally have to deal with (`NSAttributedString`a anyone?). The idea of this approach is to define our custom elements as `@IBDesignable`s, and have them set themselves up. 

Currently this only covers text and paragraph styles, open the `Onboarding.storyboard` and try changing the text style name there.

Comments and suggestions welcome!

### DIFF
--- a/Modules/Assets/Assets/Components/TextStyles.swift
+++ b/Modules/Assets/Assets/Components/TextStyles.swift
@@ -1,0 +1,128 @@
+import Foundation
+import UIKit
+
+ public struct TextStyle {
+    let style: UIFont.TextStyle
+    let size: Int
+    let weight: UIFont.Weight
+    let letterSpacing: Int
+    let lineHeight: Int
+    
+    init(style: UIFont.TextStyle,
+         size: Int,
+         weight: UIFont.Weight,
+         letterSpacing: Int,
+         lineHeight: Int) {
+        self.style = style
+        self.size = size
+        self.weight = weight
+        self.letterSpacing = letterSpacing
+        self.lineHeight = lineHeight
+    }
+    
+    func font() -> UIFont {
+        return UIFontMetrics(forTextStyle: style)
+                .scaledFont(for: UIFont.systemFont(ofSize: CGFloat(size), weight: weight))
+    }
+    
+    func paragraphStyle() -> NSParagraphStyle {
+        let paragraphStyle = NSMutableParagraphStyle()
+        paragraphStyle.lineSpacing = CGFloat(lineHeight - size)
+        return paragraphStyle
+    }
+    
+    func stringAttributes() -> [NSAttributedString.Key: Any] {
+        return [.font: font()] // will add the paragraph styles when we get proper line heights
+    }
+}
+
+public struct TextStyles {
+    public static let pageTitle = TextStyle(style: .title1, size: 24, weight: .bold, letterSpacing: 0, lineHeight: 0)
+    public static let header = TextStyle(style: .headline, size: 15, weight: .heavy, letterSpacing: 0, lineHeight: 0)
+    public static let subheader = TextStyle(style: .subheadline, size: 13, weight: .bold, letterSpacing: 0, lineHeight: 0)
+    public static let bodySmall = TextStyle(style: .body, size: 15, weight: .medium, letterSpacing: 0, lineHeight: 0)
+    public static let bodyLarge = TextStyle(style: .body, size: 17, weight: .medium, letterSpacing: 0, lineHeight: 0)
+    public static let ghostText = TextStyle(style: .body, size: 17, weight: .medium, letterSpacing: 0, lineHeight: 0)
+    public static let label = TextStyle(style: .callout, size: 12, weight: .medium, letterSpacing: 0, lineHeight: 0)
+    public static let pill = TextStyle(style: .callout, size: 15, weight: .medium, letterSpacing: 0, lineHeight: 0)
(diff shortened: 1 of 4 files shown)


juanlaube:
Some thoughts on this:

I love the idea of having predefined styles, however I would prefer the `TextStyle` to be a struct with some properties like font family, size, resolving dynamic text size, number of lines, etc, and then have some static vars with the actual styles, this gives us the flexibility to override some styles, for example I could see us getting an style and then overriding the numbers frequently.

I don't like subclassing `UILabel`, mostly because the only reason I see in the PR is because of the `IBDesignable` thing and that never works as expected, but also because some labels are not created by us (e.g: a button title, a simple cell's text, etc), instead I would add an extension method to `UILabel` to configure it with a given style.

juanlaube:
@dshr some 🤯 idea, instead of subclassing `UILabel`, I think you can add `IBInspectable` properties using extensions! Of course you can't add stored properties, but you can configure everything using get/set. I wouldn't expect `IBDesignable` to work though (if it doesn't with a regular subclass, there's no reason for it to work with an extension 😄) but you can try anyway.

Modules/Assets/Assets/Components/TextStyles.swift
```diff
@@ -0,0 +1,128 @@
...
+        public static let active = TextStyle(style: .body, size: 15, weight: .medium, letterSpacing: 0, lineHeight: 0)
+    }
+    
+    public enum Reports {
+        public static let pieChart = TextStyle(style: .callout, size: 10, weight: .semibold, letterSpacing: 0, lineHeight: 0)
+        public static let billable = TextStyle(style: .headline, size: 40, weight: .regular, letterSpacing: 0, lineHeight: 0)
+        public static let total = TextStyle(style: .headline, size: 34, weight: .regular, letterSpacing: 0, lineHeight: 0)
+        public static let amount = TextStyle(style: .subheadline, size: 28, weight: .regular, letterSpacing: 0, lineHeight: 0)
+        public static let percent = TextStyle(style: .subheadline, size: 28, weight: .regular, letterSpacing: 0, lineHeight: 0)
+    }
```
Odrakir, line 63: Do we need so specific styles? Can't this take their style from general styles?

dshr:
@juanlaube 
> I love the idea of having predefined styles, however I would prefer the `TextStyle` to be a struct with some properties like font family, size, resolving dynamic text size, number of lines, etc, and then have some static vars with the actual styles, this gives us the flexibility to override some styles, for example I could see us getting an style and then overriding the numbers frequently.

`TextStyle` is indeed a struct atm, but all the properties are `let`. We can change them to `vars`, which would then enable the overriding you wrote about. The thing is, the whole idea of these was to avoid having to override anything in out views/view controllers, since our designs will always use the styles as listed in the spec.

>I don't like subclassing UILabel, mostly because the only reason I see in the PR is because of the IBDesignable thing and that never works as expected, but also because some labels are not created by us (e.g: a button title, a simple cell's text, etc), instead I would add an extension method to UILabel to configure it with a given style.

Idk, `@IBDesignable` worked quite marvelously here, with no problems. I tried using extensions, but we'd need the stored `TextStyle` for later, when we'd integrate with other `@IBInspectable` properties, like doing smart localisation like [here](https://medium.com/@mario.negro.martin/easy-xib-and-storyboard-localization-b2794c69c9db). I guess I could look into a more hacky way to store the text style in some other field...

@Odrakir 

> First, this only takes care of text styling (the title of the PR had my hopes up). There'll be more things that we will want to style

Yeah, this is preliminary, but the general idea would still apply to all other UI elements — have a set of defined style structs that get set via IB and then applied during runtime

> This does nothing for labels with multiple styles (the main reason we used NSAttributedString) I think

Yeah, those are special cases that we'll still need to take care in code — different languages have different word orders, so this all will have to be handled the old way. Still, we have quite a lot of cases where trivial things like line spacing are custom, which also requires attributed strings.

> Then there's my comment on having too specific styles... maybe the app shouldn't have so many different styles at all. Maybe the new designs fix this

I followed the specs from design, we could probably merge some of them into one.

> Having to type the style in IB (as opposed to select it from a list) is error prone, which diminishes a bit the advantage of being able to do this in IB.

Yeah, this is not ideal, but so far `@IBInspectable` doesn't support enums. I guess we could replace it with an int selector, and then have a propertly next to it that displays the selected value as a string 🤔

> An alternative way of doing things might be to style in code via functions like this. We can still make our views IBDesignable so we can see the actual styles in IB, even if we need to edit them in code. The advantage is the code is simpler, it handles all types of views and it's type safe.

This looks neat! We can def use this to simplify the styling code. I'm not quite sure we'd mix styling in code with `@IBDesignables` unless we use custom subclasses tho..

Odrakir:
You are right about IBDesignable... it's disappointing Apple doesn't provide a way to have styles that can be easily used both in code and in IB (like with colors or assets)

I think I still like better the idea of doing this in code, even if it means we can't see the actual styles in IB (unless our components are inside a IBDesignable view). I think it'll be easier to review and merge and all we lose is the ability to see the previews in IB, which is not a big deal in my opinion as most text we show is dynamic anyway.

We could still do a subclass to show it in IB for the views the really really need it (if any).

I would like to hear more thoughts, though.

dshr:
Closing this for now, will implement the suggestions and have another go

PR #224, opened 2020-05-12
Make database tests work again

### LINKED ISSUE #164: Get Database tests running
## Description
Database tests as they are right now don't work. You can execute them manually, but Xcode or Bitrise won't find them since they are swift generic classes and those are not compatible with `XCTestCase`

So... that means they have to be refactored some how so they work. Your mission, should you choose to accept it, is exactly that.

## Definition of done
- [ ] Exactly the same tests that can be now manually run are automatically run 
- [ ] Those tests run in Xcode (when hitting command + u) and in Bitrise

juanlaube:
## Info
This makes the database tests work (again).

### Summary
The database tests didn't work because XCTest cannot discover instantiate Swift generic classes at runtime, it uses the ObjC runtime for that.

### Technical considerations
This solution moves the tests cases to new classes and adds a new class that can be instantiated at runtime for each one of the original class, then programmatically instantiates and executes all the tests cases.

It's the simplest way to solve the problem.

### Relationships
Closes #164 

### Pull Request checklist

### Pointers
Not much I can add here, it's just a workaround using some XCTest magic to make things work.

### Review template

- [ ] The code considers i18n

## Merge permissions

🦑


juanlaube:
That's a good idea, I removed the additional classes and made things work from the `DatabaseTests` class.

juanlaube:
For some reason this is failing in Bitrise and I couldn't figure out why yet.

Odrakir:
It fails locally to. [Here](https://github.com/toggl/ios/blob/aa4615352096e58bbe235b35c9aef46eaf6142d1/Modules/Database/DatabaseTests/EntityDatabaseTests.swift#L153) with a "Could not merge changes." error.

To fix it add this in line 43
```
container.viewContext.mergePolicy = NSMergeByPropertyObjectTrumpMergePolicy
```

juanlaube:
@Odrakir that doesn't seem to solve the issue, and for some reason the tests do not fail for me locally 🤔

Odrakir:
If you delete the app from the simulator I think you'll see tests fails... I think there's something wrong with the in memory DB... I'm researching it.

heytherewill:
Is anything blocking this PR? How can we unblock it?

ping @juanlaube

juanlaube:
I don't know what else to try so I'm waiting for enlightenment or for @Odrakir to find a solution, whatever comes first 😄

Odrakir:
I didn't know you were expecting me to look more into this... I'm currently out of ideas, but since this is not really urgent we can let it rest a bit more.

As a last resort, as I said on Slack, we could make the Database a dynamic framework (and move the model back from assets)

Odrakir:
Juan, I think the best thing to do for now is making this a dynamic framework and moving the model file back to it from Assets. Do you want me to do it or do you want to do it?

Odrakir:
Yay! The problem was flushing the data from the in memory database in tests. I opted for deleting the persistent store completely.

Odrakir:
Done. In the end this had to do with the tests, switching the the main thread before blocking the observable fixed it.

PR #225, opened 2020-05-12
Handle time entry autocomplete suggestions

### LINKED ISSUE #188: Update the StartEditReducer's AutocompleteSuggestionTapped action to handle TimeEntry suggestions
## Description
When the user picks a suggested a time entry, we change the details of the entry being edited/created to match the selected suggestion
Bear in mind that this method should never alter the id(s) of the running time entry, just all all other fields
## Definition of done

heytherewill:
### Summary
This makes the start edit reducer handle time entry suggestions

### Technical considerations
I renamed the autocomplete suggestion type because it had a needless hungarian suffix.
Then I changed the tag reference in the time entry so it's no longer nullable (no reason for it to be)
Then I added the needed fields to the editable time entry
et voilà

Let me know if you disagree with any of the changes

### Relationships
Closes #188 

### Pull Request checklist

### Pointers
Go at it commit by commit

### Review template
- [ ] The code considers i18n

## Merge permissions
Anyone, anytime


Modules/Database/Database/Models/ModelExtensions/TimeEntry+CoreDataModel.swift
```diff
@@ -10,7 +10,8 @@ extension TimeEntry: CoreDataModel {
             start: managedObject.start,
             duration: managedObject.duration?.doubleValue,
             billable: managedObject.billable,
-            workspaceId: managedObject.workspace.id
+            workspaceId: managedObject.workspace.id,
+            tagIds: managedObject.tags.map { $0.id }
```
Odrakir, line 14: Shouldn't you also be adding the tags in the encode method?

Modules/Database/DatabaseTests/TimeEntryDatabaseTests.swift
```diff
@@ -11,24 +11,45 @@ class TimeEntryDatabaseTests: EntityDatabaseRxExtensionsTests<TimeEntry> {
...
+            TimeEntry(id: 4, description: "TimeEntry 4", start: now, duration: 10, billable: false, workspaceId: 0, tagIds: [])
         ]
     }
 
     override var updatedEntities: [TimeEntry] {
         [
-            TimeEntry(id: 0, description: "Updated name", start: now, duration: 10, billable: false, workspaceId: 0),
-            TimeEntry(id: 1, description: "TimeEntry 1", start: now, duration: 100, billable: false, workspaceId: 1),
-            TimeEntry(id: 2, description: "TimeEntry 2", start: now, duration: 10, billable: false, workspaceId: 0)
+            TimeEntry(
```
Odrakir, line 29: Why these have each param in one line and the above don't?

Modules/Models/Models/TimeEntry.swift
```diff
@@ -11,22 +11,24 @@ public struct TimeEntry: Entity, Equatable {
...
     
     public init (
         id: Int64,
         description: String,
         start: Date,
         duration: Double?,
         billable: Bool,
-        workspaceId: Int64
+        workspaceId: Int64,
+        tagIds: [Int64]
```
Odrakir, line 23: Maybe have an empty array as a default value?

Modules/Repository/Repository/Repository.swift
```diff
@@ -107,7 +107,9 @@ extension Repository: TimeLogRepository {
                 start: time.now(),
                 duration: nil,
                 billable: false,
-                workspaceId: timeEntry.workspaceId)
+                workspaceId: timeEntry.workspaceId,
+                tagIds: []
```
Odrakir, line 111: The tag ids should come from the `StartTimeEntryDto` when starting a new TE

Modules/Timer/Timer/Common/EditableTimeEntry.swift
```diff
@@ -9,6 +9,9 @@ public struct EditableTimeEntry: Equatable {
     public var billable: Bool
     public var start: Date?
     public var duration: TimeInterval?
+    public var projectId: Int64?
+    public var taskId: Int64?
+    public var tagIds: [Int64]
     public var editableProject: EditableProject?
```
Odrakir, line 15: Unrelated to this PR, but we should rethink this a bit, this is getting messy, maybe having all TE data in its own struct (and I still think the EditableProject? should be somewhere else)

Modules/Timer/Timer/StartEdit/StartEditReducer.swift
```diff
@@ -33,7 +33,31 @@ func createStartEditReducer(repository: TimeLogRepository, time: Time) -> Reduce
...
+                fatalError()
+            case .taskSuggestion:
+                fatalError()
+            case .tagSuggestion:
+                fatalError()
+            case .createProjectSuggestion:
+                fatalError()
+            case .createTagSuggestion:
+                fatalError()
+            }
```
Odrakir, line 57: Maybe all this could go into an extension on `EditableTimeEntry`?

Modules/Database/DatabaseTests/TimeEntryDatabaseTests.swift
heytherewill: Max line size in the linter exploded after the changes added by this PR

Modules/Models/Models/TimeEntry.swift
heytherewill: My reasoning was that no other entity has a default parameter

Odrakir: No other entity has an array as a parameter

Modules/Database/DatabaseTests/TimeEntryDatabaseTests.swift
Odrakir: The ones in `initialEntities` are longer.

heytherewill: I had to change them as well

PR #226, opened 2020-05-13
Create the WheelDurationInputView

### LINKED ISSUE #149: Create the WheelDurationInputView
## Description
We need to port the wheel duration input to swift to compose the Wheel™.
While at it, let's rename it to keep the same name on both platforms where possible.

The duration input view:
- Draws a duration in "edit mode" and "display mode"
- Handles touch interaction to let the user manually input minutes
- Provides interfaces to control the widget and its represented value

## Dependencies
None.

## Definition of done
Drawing:
- [x] The input duration view draws the duration set to when in display mode
- [x] The input duration view draws the current duration in minutes when the user is inputting values

Input handling:
- [x] It's possible obtain focus when clicking on the input view
- [x] It's only possible to input digits and they work like the[ current implementation](https://github.com/toggl/mobileapp/blob/develop/Toggl.iOS/Views/EditDuration/DurationField.cs).

sofokli1:
### Summary
Adds the wheel duration input view and hooks it to the edit VC.

### Technical considerations
It is mostly ported from the old app.

Since it was working fine in the old app and it was not touched much I went for mostly 1 to 1 porting and not for a major refactor of the component.

### Relationships
Closes #149

### Pull Request checklist

### Pointers
run the app and take a look at the code

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


Modules/Timer/Timer/StartEdit/Views/DurationTextField/DurationTextField.swift
```diff
@@ -0,0 +1,120 @@
...
+            tryUpdate(input.pop())
+        } else if let number = Int(string), number >= 0 && number <= 9 {
+            tryUpdate(input.push(digit: number))
+        }
+
+        return false
+    }
+}
+
+public extension Reactive where Base: DurationTextField {
```
juanlaube, line 106: Just some food for thought: I'd rather have a simple components that do not depend on Rx, in this case for example it's already a UIControl and you could emit `.valueChanged` when setting the duration. It can be used after with RxCocoa.

Modules/Utils/Utils/Extensions/TimeInterval+Extensions.swift
```diff
@@ -12,4 +12,18 @@ extension TimeInterval {
     public func formattedDuration() -> String {
         return TimeInterval.formatter.string(from: self)!
     }
+
+    public static func from(hours: Int) -> TimeInterval {
+        TimeInterval(hours * 60 * 60)
```
juanlaube, line 17: Consider moving these magic numbers to constants, in the foreground view PR (not merged yet) I added some extensions to `Int` like `minutesInAnHour` and `secondsInAMinute`, you could use the same.

Modules/Timer/Timer/StartEdit/Views/DurationTextField/DurationTextField.swift
Odrakir: The problem with that is that then if you wan to hook it up with Rx you'll have to do some mapping.

```
wheelDurationLabelTextField.rx.value
      .map{ _ in wheelDurationLabelTextField.duration }
```

vs 
```
wheelDurationLabelTextField.rx.duration
```

sofokli1:
here are the issues for the above mentioned concerns
https://github.com/toggl/ios/issues/239
https://github.com/toggl/ios/issues/240

PR #237, opened 2020-05-15
Make the UpdateAutocompleteSuggestionsEffect return time entry suggestions

### LINKED ISSUE #187: Make the UpdateAutocompleteSuggestionsEffect return time entry suggestions
## Description
We need to update the effect that calculates autocomplete suggestions to make it handle time entry suggestions.
What we need to do is mimic the way the Xamarin app does the search. We basically split the query by spaces and make sure that every word matches some part of the time entry. This includes the description, project names, task names, tags and clients. So for example, the query `wo mob` would match an entry `Working` with the project `mobile app`, and would also match a time entry with just the description `aaasdadasmobdasdasdsawoaowowodadasdmo`

This case is to be handled last in the effect, whenever we can't match something more specific like projects/tasks or tags.

## Definition of done
- [ ] The effect is updated to return a list of time entry autocomplete suggestions
- [ ] Tests ensure that it matches the correct TEs

dshr:
### Summary
This PR fleshes out the UpdateAutocompleteSuggestionsEffect to return time entry suggestions by searching for them in the state's `entities` dict. This PR started some discussions re exactly how deterministic our state should be, please find them in the `mobile-discussion` repo.

### Relationships
Closes #187

### Pull Request checklist

### Pointers
Testing this in the runtime is rather difficult, so please go by tests and commit by commit — the last commit shuffles the files structure a bit for the sake of 💄, but makes reviewing quite confusing.

### Review template

## Merge permissions
All yours!


dshr:
@heytherewill matched the droid impl, have another look

heytherewill:
The test suite is really weak. It doesn't cover common cases liek multi word searches and partial matching happening in different places (like matching a word in the description and other word in a tag, for example). The fact that the underlyin implementation completely changed behavior and no tests broke is reason enough for us to write more tests. The Android test suite is exhaustive enough and you can get inspiration from there.

PR #241, opened 2020-05-19
Add the StopButtonTapped action

### LINKED ISSUE #156: Add the StopButtonTapped action to the StartEdit reducer
## Description
This action simply sets the duration of the `editableTimeEntry` to the current time in the StartEdit state.

## Dependencies
Depends on #142  

## Definition of done
- [x] The action is added
- [x] Tests are added
- [x] The action is dispatched from the UI
- [x] The [documentation](https://www.notion.so/toggl/StopButtonTapped-5b5d467f0cae47b8be7be83426627759) is updated to reflect the changes done in this PR

dshr:
### Summary
This PR adds the .stopButtonTapped action to our StartEditReducer

### Technical considerations
The action sets the duration only if the TE doesn't have one, otherwise it opens the end time picker.

Since we're using the same button to stop TEs and open the picker, the logic deciding when to do what had to live somewhere. For better testability I decided to have it in the reducer.

### Relationships
Closes #156

### Pull Request checklist

### Pointers
Commit by commit, folks

### Review template

## Merge permissions
Feel free


Modules/Timer/Timer/StartEdit/StartEditReducer.swift
```diff
@@ -24,6 +24,15 @@ func createStartEditReducer(repository: TimeLogRepository, time: Time) -> Reduce
         case .doneButtonTapped:
             return doneButtonTapped(state, repository)
 
+        case .stopButtonTapped:
+            if state.editableTimeEntry != nil,
+                state.editableTimeEntry!.start != nil,
+                state.editableTimeEntry!.duration == nil {
```
Odrakir, line 30: Can `state.editableTimeEntry` be nil? Shouldn't that be a fatalError?

Also... 🤔 In what case is `start` nil? I'm guessing it's the case for a new TE or a group, but then, does this action make any sense at all? I'm trying to remember how is this supposed to work. The TE won't start until the user presses Done, but what do we do with the start time?

Modules/Timer/Timer/StartEdit/StartEditReducer.swift
```diff
@@ -24,6 +25,28 @@ func createStartEditReducer(repository: TimeLogRepository, time: Time) -> Reduce
         case .doneButtonTapped:
             return doneButtonTapped(state, repository)
 
+        case .stopButtonTapped:
+            guard state.editableTimeEntry != nil else {
```
juanlaube, line 29: With something like this you will avoid all the force unwrapping that follow:
```suggestion
            guard let editableTimeEntry = state.editableTimeEntry else {
```

Modules/Timer/Timer/StartEdit/StartEditReducer.swift
```diff
@@ -24,6 +25,28 @@ func createStartEditReducer(repository: TimeLogRepository, time: Time) -> Reduce
...
             return doneButtonTapped(state, repository)
 
+        case .stopButtonTapped:
+            guard let editableTimeEntry = state.editableTimeEntry else {
+                fatalError("Trying to stop time entry when not editing a time entry")
+            }
+            
+            if editableTimeEntry.start == nil {
+                state.editableTimeEntry!.start = time.now()
+                state.editableTimeEntry!.duration = 0
```
juanlaube, line 35:
```suggestion
                editableTimeEntry.start = time.now()
                editableTimeEntry.duration = 0
```

Modules/Timer/Timer/StartEdit/StartEditReducer.swift
```diff
@@ -24,6 +25,28 @@ func createStartEditReducer(repository: TimeLogRepository, time: Time) -> Reduce
...
+                return []
+            }
+            
+            if editableTimeEntry.duration == nil {
+                let maxDuration = TimeInterval.maximumTimeEntryDuration
+                let duration = time.now().timeIntervalSince(editableTimeEntry.start!)
+                if duration > 0 && duration <= maxDuration {
+                    state.editableTimeEntry!.duration = duration
```
juanlaube, line 43:
```suggestion
                    editableTimeEntry.duration = duration
```

dshr:
> I meant something like this 😄

It’s going to be a `let` constant then, can’t modify it 🤷🏻‍♂️

PR #246, opened 2020-05-20
Add the date picker to the edit duration view

### LINKED ISSUE #151: Add the wheel widget
## Description
The Wheel™ is going to be ported almost 1:1 from the Xamarin apps (api differences apply).
The Wheel™ is composed of the following components:
- The [WheelBackgroundView](https://github.com/toggl/mobileapp/blob/develop/Toggl.iOS/Views/EditDuration/WheelBackgroundView.cs) which handles drawing the fixed background of the wheel (clockdial and empty wheel background).
- The [WheelForegroundView](https://github.com/toggl/mobileapp/blob/develop/Toggl.iOS/Views/EditDuration/WheelForegroundView.cs) (the main component that handles most of the user interaction, drawing and exposes most of the API used to control the widget).
- The [WheelDurationInputView](https://github.com/toggl/mobileapp/blob/develop/Toggl.iOS/Views/EditDuration/DurationField.cs) which handles manual duration input in minutes.

## Dependencies
It needs the base layout that should be added by #142  
Wheel Background View #148  
Wheel Foreground View #150  
Wheel Input Duration View #149 

## Definition of done
- [ ] The WheelBackgroundView is added
- [ ] The WheelForegroundView is added
- [ ] The WheelDurationInputView is added
- [ ] The components are combined into one and the placeholder added by #142 is replaced with the actual Wheel™

### LINKED ISSUE #159: Handle displaying the datetime barrels for the edit duration view
## Description
We need to display the datetime barrels when the StartEdit's state say so.

## Dependencies
Depends on #142  
Depends on #143  
Depends on #145  

## Definition of done
- [x] A datetime picker (barrel) is displayed when the `StartEdit`'s state `datetimePickMode` is set to `Start` or `End`.
- [x] The barrel has its max and min datetime set based on the `editableTimeEntry`'s start time, end time (start time + duration) and the `dateTimePickMode`.
- [x]  The preset datetime is set based on the `datetimePickMode` and the `editableTimeEntry`'s `startTime` and `duration`.
- [x] No picker should be visible when `datetimePickMode` is set to `None`.
- [x] The `DateTimePickingCancelled` action is dispatched when label that opened the barrels is clicked again.

### LINKED ISSUE #160: Dispatch DateTimePicked actions when picking datetimes
## Description
We need to dispatch `DateTimePicked` actions when datetimes are picked.

## Dependencies
Depends on #159 

## Definition of done
- [ ] When the barrel picks a datetime, a `DateTimePicked` action is dispatched with its value

juanlaube:
## Info
This adds the date picker to the edit duration view

### Summary
This adds the date picker to the edit duration view and refactor the actions to set the date picker mode so the logic is in the reducer.

### Technical considerations
I did a couple of extra things in this PR:
- Moved the wheel views into a container view
- Added the calculation of min and max start and stop dates for the editable time entry in an extension

### Relationships
Closes #151 
Closes #159
Closes #160 

### Pull Request checklist

### Pointers
Not much I can add other than testing it.

### Review template
- [ ] The code considers i18n

## Merge permissions
🦑


Odrakir:
Before going deep into code review... I think the `StartEditViewController` is getting too big and too messy, it has many responsibilities and many components coexist. 

I think it might be worth exploring using the table view as initially planned. Meaning the wheel and the date picker are different kind of cells in the table view (more `StartEditCellType` cases). We could make the cell height be constrained based (with `UITableView.automaticDimension`), this way we just have to add the animation code inside the cell and not in the view controller.

I'm open to alternatives, but we need to find a way to deal with that code now.

PR #247, opened 2020-05-20
 Make the UpdateAutocompleteSuggestionsEffect return project suggestions

### LINKED ISSUE #189: Make the UpdateAutocompleteSuggestionsEffect return project/task suggestions
## Description
We need to update the effect that calculates autocomplete suggestions to make it handle project and task suggestions.
What we need to do is mimic the way the Xamarin app does the search. We find if the caret is currently placed in a word that starts with an @. If it is, we use that word to suggest projects and tasks, matching against the project/task name. Check the Xamarin implementation for more details
The list of suggestions returned must always include a `CreateProject` autocomplete suggestion, with the word that's included in the @ as the suggested name.

## Definition of done
- [ ] The effect is updated to return a list of project/task/create project autocomplete suggestions

dshr:
### Summary
This PR adds the query symbol detection code and handles the case where we look for projects.

### Relationships
Closes #189 

### Pull Request checklist

### Pointers
Commit by commit

### Review template

## Merge permissions
All yours!


Modules/Timer/Timer/StartEdit/Effects/UpdateAutocompleteSuggestionsEffect.swift
```diff
@@ -4,20 +4,47 @@ import Models
...
                                          _ position: Int) -> [Effect<StartEditAction>] {
     guard let query = state.editableTimeEntry?.description
         else { fatalError("No editable time entry while looking for autocomplete suggestions") }
+    let result = Single.just(getSuggestions(for: query, at: position, in: state.entities))
+        .map(StartEditAction.autocompleteSuggestionsUpdated)
+        .toEffect()
+    return [result]
+}
+
+func getSuggestions(for query: String, at position: Int, in entities: TimeLogEntities) -> [AutocompleteSuggestion] {
```
Odrakir, line 21: private

Modules/Timer/Timer/StartEdit/Effects/UpdateAutocompleteSuggestionsEffect.swift
```diff
@@ -52,3 +79,20 @@ func searchTimeEntries(for words: [String], in entities: TimeLogEntities) -> [Ti
         leftHand.description > rightHand.description
     })
 }
+
+func searchProjects(for words: [String], in entities: TimeLogEntities) -> [Project] {
```
Odrakir, line 83: private

Modules/Timer/Timer/StartEdit/Effects/UpdateAutocompleteSuggestionsEffect.swift
```diff
@@ -4,20 +4,47 @@ import Models
...
-    return [
-        Single.just(suggestions)
-            .map(StartEditAction.autocompleteSuggestionsUpdated)
-            .toEffect()
-    ]
+    let timeEntries = searchTimeEntries(for: words, in: entities)
+    return timeEntries.map(AutocompleteSuggestion.timeEntrySuggestion)
 }
 
 func searchTimeEntries(for words: [String], in entities: TimeLogEntities) -> [TimeEntry] {
```
Odrakir, line 50: private

Modules/Timer/Timer/StartEdit/Effects/UpdateAutocompleteSuggestionsEffect.swift
```diff
@@ -4,20 +4,47 @@ import Models
 import RxSwift
 import Repository
 
+private let querySymbols: [Character] = ["@"]
+
 func updateAutocompleteSuggestionsEffect(_ state: StartEditState,
                                          _ repository: TimeLogRepository,
                                          _ description: String,
```
Odrakir, line 11: I don't think `description` is used anymore. On the other hand the state is decomposed immediately in `description` and `entities`. Wouldn't it be easier if this method just got a `description: String` and `entities: TimeLogEntities`?

dshr: Yeah, that does make sense. In all fairness, the fact that we have `repository` there confused me to bits when doing to TE version of this.

dshr:
> This crashes for me as soon as I see the timeline. Before I tap the textfield or anything. I think the descriptionEntered action is getting dispatched by default when everything is created, which probably shouldn't. Can you check that?

Works fine for me 🤷‍♂️ . What is the crash message?

Odrakir:
It crashes [here](https://github.com/toggl/ios/pull/247/files#diff-20ac8a4e331b35ef7d9d6550a35c2920R13) with "Fatal error: No editable time entry while looking for autocomplete suggestions".

As I said this is before I bring the start view up, I think the descriptionEntered shouldn't fire.

PR #253, opened 2020-05-21
Project suggestions

### LINKED ISSUE #190: Update the StartEditReducer's AutocompleteSuggestionTapped action to handle Project suggestions
## Description
When you select a project suggestion, it sets the project of the editable time entry.
Selecting a project suggestion should also remove the word that starts with an @ from the description (do so by checking the cursor position as well)

## Definition of done

heytherewill:
### Summary
This implements a proper search and replace behavior (as per https://github.com/toggl/android/pull/320) for autocomplete suggestions in order to properly handle project suggestions

### Technical considerations
This is the proper use case that covers all corner cases. Feel free to suggest changes to the Swift code, since, well, I'm not the most proficient at it.

### Relationships
Closes #190 

### Pull Request checklist

### Pointers
Review per commit. Check the tests to see what the expected behavior is and if I covered all corner cases. Check if I handled substrings properly as well.

### Review template

## Merge permissions
Anyone


Modules/Timer/Timer/StartEdit/Extensions/String+Extensions.swift
```diff
@@ -0,0 +1,56 @@
...
+            let matchedTheFirstWord = tokens.contains { matchSubstring.starts(with: String($0)) }
+            let queryWithToken = String(matchedTheFirstWord ? matchSubstring : matchSubstring.dropFirst())
+            
+            let token = queryWithToken.first
+            let query = String(queryWithToken.dropFirst())
+            
+            return (token, query)
+
+        } catch {
+            return (nil, "")
```
dshr, line 37: Maybe return the original string instead?
```suggestion
            return (nil, self)
```

PR #257, opened 2020-05-22
Local state and optional state

Odrakir:
### Summary
This is a draft so we can further discussed what's described [here](https://github.com/toggl/mobile-discussions/issues/65).

### Technical considerations
The PR basically involves three things, if you read the commits one by one is easy to understand:
- Change the way we deal with local state by using a private backing property
- Allow for optional state and make StartEditState optional, then use the `optional` operator on reducers.
- Make the view controller or coordinator of the start edit view deal with this. THIS IS STILL PENDING.

The optional part is taken from [here](https://github.com/pointfreeco/swift-composable-architecture/blob/master/Examples/CaseStudies/SwiftUICaseStudies/01-GettingStarted-OptionalState.swift). It's SwiftUI so they deal with the view being optional with a `IfLetStore`, which we can't do, so that's the pending part, but it wouldn't be much complicated to do something equivalent for UIKit.

NOTE: The PR as it is right now doesn't work! (it needs the 3rd point described up there)

The crash happens [here](https://github.com/toggl/ios/pull/257/files#diff-b3ab72517359c1f70e93b5e63242567aR45) because that state can be nil (and it's nil at the beginning). So we need to find a way so it only attempts to "view" that store when that piece of state is not nil.

## Merge permissions
Don't. This is just a draft for discussion.

### DIFF
--- a/Modules/Architecture/Architecture/Core/Reducer.swift
+++ b/Modules/Architecture/Architecture/Core/Reducer.swift
@@ -34,6 +34,13 @@ public struct Reducer<StateType, ActionType> {
     ) -> Reducer<StateType, GlobalAction> {
         return self.pullback(state: \StateType.self, action: action)
     }
+
+    public var optional: Reducer<StateType?, ActionType> {
+        .init { state, action in
+            guard state != nil else { return [] }
+            return self.reduce(&state!, action)
+        }
+    }
 }
 
 public func combine<State, Action>(
--- a/Modules/Onboarding/Onboarding/Common/OnboardingState.swift
+++ b/Modules/Onboarding/Onboarding/Common/OnboardingState.swift
@@ -3,42 +3,15 @@ import Models
 import Architecture
 import Utils
 
-public struct LocalOnboardingState: Equatable {
-    internal var email: String = ""
-    internal var password: String = ""
-    
-    public init() {
-    }
-}
-
 public struct OnboardingState: Equatable {
     public var user: Loadable<User>
     public var route: RoutePath
-    public var localOnboardingState: LocalOnboardingState
-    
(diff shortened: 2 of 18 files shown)


heytherewill:
I really like the new way of doing local states. I'm not sure how (if at all) we'd translate that to Kotlin, but it doesn't really matter since the final representation of the states that matter is still the same. Good stuff there.

The optional bit also looks good, though I'd just say be careful with making the editableTimeEntry non optional; after all, you need to set it to nil to dismiss the view 🤔 

That being said, I don't like the fact that we're going back on something we agreed on, which was to not extract specific actions to be handled elsewhere. Having each reducer being exhaustive on each action it proposes to handle is a really good benefit in terms of readability and it makes browsing code so much simpler. If we don't enforce that with an iron fist, in the future we might have actions scattered all throughout the codebase and that's not cool. 

If a particular reducer is too big because it handles too many actions, we can break it into more modules. One more module on its own is not enough to increase complexity more than a 1000 line file. 
If a particular reducer is too repetitive because its code is duplicated, we can extract reducers to their own files in order to simplify the reducer code.

Odrakir:
Yeah. We are on the same page. My idea at first was that moving code to higher reducers within the same feature wouldn’t be that bad. But you are right, it’s not ideal. 

I’ll try to fix that on the draft. 🤔 I’m thinking there might be another option directly supported by the architecture.

Odrakir:
Also, the whole idea behind making `EditableTimeEntry` non optional is to avoid all the guards in the reducer. `StartEditReducer` always has an `EditableTimeEntry`, so the optional part in reality is `StartEditState`. The dismissal needs to happen in some other way.

Odrakir:
I'm closing this so we make changes one at a time.

PR #258, opened 2020-05-22
Handle Task and tag suggestions

### LINKED ISSUE #190: Update the StartEditReducer's AutocompleteSuggestionTapped action to handle Project suggestions

### LINKED ISSUE #198: Update the StartEditReducer's AutocompleteSuggestionTapped action to handle tag suggestions
## Description
When you select a tag suggestion, it adds the tag to the list of tags of the editable time entry.
Selecting a tag suggestion should also remove the word that starts with an # from the description (do so by checking the cursor position as well)

## Definition of done

heytherewill:
### Summary
This handles our tags and task suggestions

### Technical considerations
I did both in a single PR because the heavy lift for both is the same and the actual changes become 3 liners, so might as well do it all at once

### Relationships
Closes #190 
Closes #198 

### Pull Request checklist

### Pointers
Review per commit

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone anytime


Modules/Timer/Timer/StartEdit/StartEditReducer.swift
```diff
@@ -91,12 +91,12 @@ func createStartEditReducer(repository: TimeLogRepository, time: Time) -> Reduce
             
         case .projectButtonTapped, .addProjectChipTapped:
             guard let editableTimeEntry = state.editableTimeEntry else { return [] }
-            state.editableTimeEntry?.description = appendCharacter("@", toString: editableTimeEntry.description)
+            state.editableTimeEntry?.description = appendCharacter(projectToken, toString: editableTimeEntry.description)
```
juanlaube, line 94: The build fails because `appendCharacter`, against its name, takes a `String` as the first parameter.

Modules/Timer/Timer/StartEdit/StartEditReducer.swift
```diff
@@ -91,12 +91,12 @@ func createStartEditReducer(repository: TimeLogRepository, time: Time) -> Reduce
...
             return []
             
         case .tagButtonTapped, .addTagChipTapped:
             guard let editableTimeEntry = state.editableTimeEntry else { return [] }
-            state.editableTimeEntry?.description = appendCharacter("#", toString: editableTimeEntry.description)
+            state.editableTimeEntry?.description = appendCharacter(tagToken, toString: editableTimeEntry.description)
```
juanlaube, line 99: Same

Modules/Timer/Timer/StartEdit/StartEditReducer.swift
```diff
@@ -189,24 +189,36 @@ extension EditableTimeEntry {
...
+            removeQueryFromDescription(tagToken, cursorPosition)
+            
         case .createProjectSuggestion:
             fatalError()
         case .createTagSuggestion:
             fatalError()
         }
     }
+    
+    mutating func removeQueryFromDescription(_ token: Character, _ cursorPosition: Int) {
```
Odrakir, line 216: According to the latest designs we are not removing the tokens from the description after selecting them.

dshr:
This closes #195, not #198

PR #259, opened 2020-05-25
Reducer decoration

### LINKED ISSUE #260: Reducer decorators
## Description
Add a way to decorate reducers with another reducer so they gain their behaviour. Base reducers should include the needed state and an action to redirect decorator actions.

## Definition of done
- [ ] Add instance method on reducer to decorate them
- [ ] Create core time entries reducer
- [ ] Decorate all reducers that handle time entries with the time entries reducer

- NOTE: Soon we will probably need to extract this reducer outside of the Timer module, so other modules can also be decorated with it (Calendar, probably)

Odrakir:
## Info
In order to simplify reducers this PR created one single reducer to handle all time entry operations (start, continue, delete, stop, modify). Then it "decorates" existing reducers with this behaviour.

I think code looks much nicer and this technique can be used for many other parts of reducers.

### Relations
Closes #260 

### Technical considerations
Other than what's mentioned above this PR also changes a couple of things when decoding TEs (there was a bug) and disables DB access for now, so we can test the app with dummy data.

### Pull Request checklist

### Pointers
Go one commit at a time, I think they are self explanatory.

### Review template

- [ ] The code considers i18n

## Merge permissions
Me


Odrakir:
That might make sense in the future. Let's start with this and see.

Modules/Timer/Timer/TimeEntriesLog/Views/TimeEntriesLogViewController.swift
```diff
@@ -43,8 +42,8 @@ public class TimeEntriesLogViewController: UIViewController, Storyboarded {
                     }
                     cell.descriptionLabel.text = item.description
                     cell.descriptionLabel.textColor = item.descriptionColor
-                    cell.projectClientTaskLabel.textColor = item.projectColor
-                    cell.projectClientTaskLabel.attributedText = item.projectTaskClient
+//                    cell.projectClientTaskLabel.textColor = item.projectColor
```
dshr, line 45: Do these need to be commented out?

Modules/Timer/TimerTests/Reducers/TimeEntriesReducerTests.swift
```diff
@@ -0,0 +1,159 @@
+import XCTest
+import Architecture
+import Models
+import OtherServices
+//import RxBlocking
```
dshr, line 5: 👀

Modules/Timer/TimerTests/Reducers/TimeEntriesReducerTests.swift
```diff
@@ -0,0 +1,159 @@
...
+//import RxTest
+@testable import Timer
+
+class TimeEntriesReducerTests: XCTestCase {
+
+    var now = Date(timeIntervalSince1970: 987654321)
+    var mockRepository: MockTimeLogRepository!
+    var mockTime: Time!
+    var reducer: Reducer<[Int64: TimeEntry], TimeEntriesAction>!
+//    var testScheduler = TestScheduler(initialClock: 0)
```
dshr, line 15: 👀

Modules/Timer/Timer/TimeEntriesLog/Views/TimeEntriesLogViewController.swift
Odrakir: Those lines are making my app crash, not sure why, but since we still have to go over the design of those cells I think we can ignore that for now.

dshr: What does the crash say? It doesn't happen to me 🤷‍♂️

PR #266, opened 2020-05-25
Make the UpdateAutocompleteSuggestionsEffect return tag suggestions

### LINKED ISSUE #193: Make the UpdateAutocompleteSuggestionsEffect return tag suggestions
## Description
We need to update the effect that calculates autocomplete suggestions to make it handle tag suggestions.
What we need to do is mimic the way the Xamarin app does the search. We find if the caret is currently placed in a word that starts with an #. If it is, we use that word to suggest tags. Check the Xamarin implementation for more details
The list of suggestions returned may include a `CreateProject` autocomplete suggestion, with the word that's included in the # as the suggested name. It will only do so if the tag does not exist in the workspace of the time entry.

## Definition of done
- [ ] The effect is updated to return a list of Tag/create tag autocomplete suggestions

dshr:
### Summary
This PR adds search by tag name to autocomplete

### Relationships
Closes #193

### Pull Request checklist

### Pointers
Commit by commit

### Review template

## Merge permissions
Feel free


Modules/Timer/Timer/StartEdit/Effects/UpdateAutocompleteSuggestionsEffect.swift
```diff
@@ -51,7 +51,18 @@ func fetchProjectSuggestions(for query: String, in entities: TimeLogEntities) ->
 }
 
 func fetchTagSuggestions(for query: String, in entities: TimeLogEntities) -> [AutocompleteSuggestion] {
-    return []
+    let words = query.split(separator: " ").map { String($0) }
+    
+    var suggestions = words.reduce(Array(entities.tags.values)) { tags, word in
```
heytherewill, line 56: Only tags in the current workspace should be used as input

Modules/Timer/Timer/StartEdit/Effects/UpdateAutocompleteSuggestionsEffect.swift
```diff
@@ -51,7 +51,18 @@ func fetchProjectSuggestions(for query: String, in entities: TimeLogEntities) ->
...
+        return tags.filter { tag in
+            return tag.name.contains(word)
+        }
+    }.sorted(by: { leftHand, rightHand in
+        leftHand.name > rightHand.name
+    }).map(AutocompleteSuggestion.tagSuggestion)
+    
+    suggestions.insert(AutocompleteSuggestion.createTagSuggestion(name: query), at: 0)
```
heytherewill, line 64: This should only ever be added if no tags match exactly the query.

Modules/Timer/TimerTests/Reducers/StartEdit/StartEditReducerTests+AutocompleteTags.swift
```diff
@@ -0,0 +1,39 @@
+import XCTest
+import Architecture
+import Models
+import OtherServices
+@testable import Timer
+
+extension StartEditReducerTests {
+    func test_autocomplete_withTagToken_returnsTagsMatchingByName() {
```
heytherewill, line 8: Add cases for checking the `createTagSuggestion` existence. Also add tests that consider the current workspace of tags.

Modules/Timer/Timer/StartEdit/Effects/UpdateAutocompleteSuggestionsEffect.swift
dshr: Didn't know this, will change the impl and tests accordingly.

PR #267, opened 2020-05-25
Add the test scheduler provider infrastructure

### LINKED ISSUE #98: Creating the test infrastructure for testing with schedulers
## Description
We need an option to move time forward so that tests do not have to wait for delay observables.

## Dependencies
#97

## Definition of done
- [ ] Add the infrastructure
- [ ] Use it in the following tests in `TimeEntriesLogReducerTests.swift`
    - [ ] `testTimeEntrySwipedLeftWhenNoOtherIdIsScheduledForDeletion`
    - [ ] `testTimeEntrySwipedLeftWhenOtherIdIsScheduledForDeletion`
    - [ ] `testTimeEntrySwipedLeftGroupWhenNoOtherIdIsScheduledForDeletion`
    - [ ] `testTimeEntrySwipedLeftGroupWhenOtherIdIsScheduledForDeletion`

juanlaube:
## Info
This adds the test scheduler provider infrastructure.

### Summary
As the title says, this introduces a test scheduler provider, what we had as SchedulerProvider is now a class that implements the same protocol as the TestSchedulerProvider.

### Relationships
Closes #98 

### Pull Request checklist

### Pointers
While this introduces the scheduler provider protocol and replaces the TestScheduler in the tests, the fast forward is not included since tests do not make use of that.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


juanlaube:
Yeah, I see your point Ricardo, I started over now without the `SchedulerProvider` protocol.
Now I added the `receiveAfter(TestTime)` step case, but I'm stuck with some tests, I need another pair of eyes here.

sofokli1:
the problem with multiple actions in one step is that all may not happen after the delay. EX: `testTimeEntrySwipedLeftWhenOtherIdIsScheduledForDeletion` has the `timeEntryDeleted` action at time 0 and the `commitDeletion` at time 5. So having the delay in as associated value in the stepType may not work in this case (works fine for single action steps though). In this kind of situation it may make more sense to have the delay encapsulated with the action. Maybe a different `assertReducerFlow` that accepts steps of type `Step<State, (Action, **TestTime**)>...`. Just a thought, maybe there is more elegant way of doing this

sofokli1:
just for fun played with the idea for a couple of minutes. this came out
```swift
func testTimeEntrySwipedLeftWhenOtherIdIsScheduledForDeletion() {

        let swipedTimeEntryId: Int64 = 0
        let waitingToBeDeletedId: Int64 = 1

        let timeEntries = createTimeEntries(now)

        var entities = TimeLogEntities()
        entities.timeEntries = timeEntries
        let state = TimeEntriesLogState(entities: entities, expandedGroups: [], entriesPendingDeletion: [waitingToBeDeletedId])

        assertReducerFlow1(
            initialState: state,
            reducer: reducer,
            testScheduler: testScheduler,
            steps:
            Step(.send, (.timeEntrySwiped(.left, swipedTimeEntryId), TestTime(0))) {
                $0.entriesPendingDeletion = [swipedTimeEntryId]
            },
            Step(.receive, [(.timeEntryDeleted(waitingToBeDeletedId), TestTime(0)), (.commitDeletion([swipedTimeEntryId]), TestTime(5))]) {
                $0.entities.timeEntries[waitingToBeDeletedId] = nil
                $0.entriesPendingDeletion.removeAll()
            },
            Step(.receive, (.timeEntryDeleted(swipedTimeEntryId), TestTime(5))) {
                $0.entities.timeEntries[swipedTimeEntryId] = nil
            }
        )
    }

func assertReducerFlow1<State: Equatable, Action: Equatable>(
    initialState: State,
    reducer: Reducer<State, Action>,
    testScheduler: TestScheduler? = nil,
    steps: Step<State, (Action, TestTime)>...,
    file: StaticString = #file,
    line: UInt = #line
) {
    var state = initialState
    var effects = [Effect<Action>]()
    steps.forEach { step in
        var expected = state

        switch step.type {

        case .send:
            if !effects.isEmpty {
                XCTFail("Action sent before handling \(effects.count) pending effect(s)", file: step.file, line: step.line)
            }
            let reducerEffects = step.actions.flatMap { reducer.reduce(&state, $0.0) }
            effects.append(contentsOf: reducerEffects)

//        case .receive:
//            guard !effects.isEmpty else {
//                XCTFail("No pending effects to receive from", file: step.file, line: step.line)
//                break
//            }
//
//            var actions = [Action]()
//            if testScheduler == nil {
//                actions = effects.compactMap { try? $0.asSingle().toBlocking().first() }
//            } else {
//                _ = Observable.from(effects.map({ $0.asSingle().asObservable() }))
//                    .merge()
//                    .toArray()
//                    .subscribe(onSuccess: { actions = $0 })
//                testScheduler!.start()
//            }
//            effects = []
//
//            XCTAssert(actions.equalContents(to: step.actions), file: step.file, line: step.line)
//            let reducerEffects = actions.flatMap { reducer.reduce(&state, $0) }
//            effects.append(contentsOf: reducerEffects)

        case .receive:
            guard !effects.isEmpty else {
                XCTFail("No pending effects to receive from", file: step.file, line: step.line)
                break
            }

            guard let testScheduler = testScheduler else {
                XCTFail("Test scheduler is not assigned")
                break
            }

            let observer = testScheduler.createObserver(Action.self)
            _ = Observable.from(effects.map({ $0.asSingle().asObservable() }))
                .merge()
                .subscribe(observer)

            testScheduler.start()

            let actions = observer.events.compactMap { $0.value.element }
            let expectedEvents: [Recorded<Event<Action>>] = step.actions.map { .next($0.1, $0.0) } + [.completed(step.actions.last!.1)]
            effects = []
            print("++")
            print(actions)
            print("--")
            print(expectedEvents)
            print("==")
            print(observer.events)

            XCTAssert(observer.events.equalContents(to: expectedEvents), file: step.file, line: step.line)
            let reducerEffects = actions.flatMap { reducer.reduce(&state, $0) }
            effects.append(contentsOf: reducerEffects)

        case .receiveAfter(_):
            break
        }

        step.update(&expected)
        XCTAssertEqual(state, expected, file: step.file, line: step.line)
    }

    if !effects.isEmpty {
        XCTFail("Assertion failed to handle \(effects.count) pending effect(s)", file: file, line: line)
    }
}
```

Odrakir:
Oh... I see what you are trying to achieve. Have you seen the latest PointFree episode? They introduce a way of dealing with schedulers the testing reducers that I think we should definitely implement.

Basically they have another type of step called `do` that allows to introduce imperative tasks that happen between normal steps.

In the example below that's used to advance the test scheduler to reflect the delay the reducer is adding.

We wouldn't need to inject the test scheduler in `assertReducerFlow` anymore.

```swift
store.assert(
  .send(.todo(index: 0, action: .checkboxTapped)) {
    $0.todos[0].isComplete = true
  },
  .do {
    self.scheduler.advance(by: 0.5)
  },
  .send(.todo(index: 0, action: .checkboxTapped)) {
    $0.todos[0].isComplete = false
  },
.do {
    self.scheduler.advance(by: 1)
  },
  .receive(.todoDelayCompleted)
)
```

Odrakir:
Also, I think we should stop returning array of effects from reducers and instead make Effects be Observables instead of Singles. This requires some changes to the store, and maybe it has to be done in another PR.

PR #268, opened 2020-05-25
Handle Create project suggestions

### LINKED ISSUE #191: Update the StartEditReducer's AutocompleteSuggestionTapped action to handle create project suggestions
## Description
When you select a create project suggestion, it sets the editableProject property and that opens the project creation popup

## Definition of done
- [ ] The project creation sheet appears when the project created autocomplete suggestion is tapped

heytherewill:
### Summary
This adds handling project creation suggestions

### Technical considerations
One of the commits makes the bottom sheet view controller generic. This is needed to actually show the view controller that creates the projects.

### Relationships
Closes #191 

### Pull Request checklist

### Pointers
Review per commit

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone, anytime


Modules/Timer/Timer/Common/Views/BottomSheetViewController.swift
```diff
@@ -112,13 +97,13 @@ class StartEditBottomSheet<ContainedView: UIViewController>: UIViewController, U
         layout()
     }
 
-    private func hide() {
+    func hide() {
```
Odrakir, line 100: Why make them public? They don't seem to be called from anywhere else. It might make sense to move the subscription somewhere else, though

Modules/Timer/Timer/Common/Views/BottomSheetViewController.swift
```diff
@@ -249,3 +234,18 @@ class StartEditBottomSheet<ContainedView: UIViewController>: UIViewController, U
...
+    func loseFocus()
+    func focus()
+    func dispatchDialogDismissed()
+}
+
+enum BottomSheetState {
+    case hidden
+    case partial
+    case full
+}
```
Odrakir, line 251: Do we have guidelines for this? When you remove them from the top I thought you were moving them to other file (which makes sense), but why change them to put them at the bottom?

heytherewill: These should indeed be other files

Modules/Timer/Timer/Common/Views/BottomSheetViewController.swift
heytherewill: Likely an oopsie

heytherewill:
@Odrakir You can test this by changing the close button to emit the correct action:

So in `StartEditViewController.swift`, line 100:
```swift
        closeButton.rx.tap
            .mapTo(StartEditAction.autocompleteSuggestionTapped(.createProjectSuggestion(name: "Test")))
            .subscribe(onNext: store.dispatch)
            .disposed(by: disposeBag)
```

PR #286, opened 2020-05-27
Handle tag creation suggestions

### LINKED ISSUE #195: Update the StartEditReducer's AutocompleteSuggestionTapped action to handle tag creation suggestions
## Description
When you select a create tag suggestion, it returns an effect that creates a new tag. This effect should then emit a TagCreated action which handles adding the tag to the editable TE.
Selecting a tag creation suggestion should also remove the word that starts with a # from the description (do so by checking the cursor position as well)

## Dependencies
Depends on #194

## Definition of done

- Tag suggestions add the tag to the list of tags of the editable time entry 
- Create tag suggestions create a new tag using an effect. This effect should then emit a TagCreated action which handles adding the tag to the editable TE

heytherewill:
### Summary
This adds code to handle tag cration suggestions

### Technical considerations
Some repository boilerplate code was needed. No biggie

### Relationships
Closes #195

### Pull Request checklist

### Pointers
Review commit by commit

### Review template

## Merge permissions
Anyone, anytime


Modules/Repository/Repository/Models/CreateTagDto.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+import Models
+
+public struct CreateTagDto {
```
Odrakir, line 4: Just one question, why `CreateTagDto` and not `TagDTO`?  I ask this related to a comment I read on Slack regarding `StartTimeEntryDTO` and `CreateTimeEntryDTO`... Do we really need all those? Isn't `TimeEntryDTO` enough?

heytherewill: For time entries, it isn't. There's a big difference between starting an entry (which requires some checks in the DB for stopping) and creating a time entry. This means that the DTO for starting shouldn't have the duration, otherwise this will give consumers of the method false expectations

Regarding tags, yeah, just a tag DTO should suffice

Regarding `Dto` over `DTO`, I prefer the later, but the project Dto already had `Dto`in its name, so I was consistent. I will just rename both :)

PR #287, opened 2020-05-27
Add default project colors

### LINKED ISSUE #255: Add default colors to projects
## Description
We need to add the default colors as a property of the companion object of the project class and use that to create editable projects with a random default color

The default colors are:

```swift
let defaultColors = ["#0B83D9", "#9E5BD9", "#D94182", "#E36A00", "#BF7000", "#C7AF14", "#D92B2B", "#2DA608", "#06A893", "#C9806B", "#465BB3", "#990099", "#566614", "#525266"]
```

The default colors are part of our business logic and, thus, shouldn't be declared as resources. 

## Definition of done
- [ ] A property is added to the companion object of the project class containing our default colors
- [ ] That property is used when creating an editable project, to assign a random color to it.

heytherewill:
### Summary
This adds Toggl's default colors and randomly assigns one of them to newly created projects

### Relationships
Closes #255 

### Pull Request checklist
- [ ] One of our [live or file custom templates](https://github.com/toggl/android/tree/master/.idea/templates) was used in this PR

### Pointers
This is so easy you can just look at the changes

### Review template

## Merge permissions
Anyone, anytime


Modules/Timer/TimerTests/Reducers/StartEdit/StartEditReducerTests.swift
```diff
@@ -594,11 +594,14 @@ class StartEditReducerTests: XCTestCase {
...
-                $0.editableTimeEntry?.editableProject = editableProject
-        })
+            Step(.send, .autocompleteSuggestionTapped(suggestion), assert: { actualState in
+                guard let editableProject = actualState.editableTimeEntry?.editableProject else { fatalError() }
+                
+                XCTAssertEqual(editableProject.workspaceId, workspaceId)
+                XCTAssertEqual(editableProject.name, projectName)
+                XCTAssertTrue(Project.defaultColors.contains(editableProject.color))
+            })
+        )
```
Odrakir, line 604: The problem with this new `assert` parameter is that the test is not as good anymore. You are just asserting a specific part of the state, where as when using `update` we are comparing the whole state the reducer handles.

The reason you can't use `update` if because there's a dependency that you are not handling which is the randomisation of the project color. Maybe we can find another way of doing that?

dshr: For a way of doing this I can suggest the general idea of this: https://www.pointfree.co/collections/randomness/predictable-randomness

PR #304, opened 2020-05-28
Add the calendar items selector

### LINKED ISSUE #202: Create the visible items selector
## Description

We need to add a selector to merge time entries and calendar events into a collection of group of calendar items. These items are grouped in a way they don't overlap with each other when rendering them in a collection view.
The selector needs to also filter the time entries and calendar events for the selected date.

The `CalendarItem` will be a merge of the old CalendarItem class and CalendarItemLayoutAttributes, its shape should be something like this:

```
struct CalendarItem {
    enum Value {
        case .timeEntry(TimeEntry)
        case .calendarEvent(CalendarEvent)
    }

    let value: Value
    let columnIndex: Int
    let totalColumns: Int
}

```

## Definition of done
- [ ] The CalendarItem struct is added
- [ ] The selector returns items for the selected date only
- [ ] The logic to group items (i.e: calculate the columns) is translated to the selector from the CalendarLayoutCalculator in the old app
- [ ] Tests are created for the selector

Note: In the old app, the layout calculator does a sort by index to keep the order stable so it can match the order of the input, this is not needed anymore.

Depends on #201

juanlaube:
## Info
This adds the calendar items selector.

### Summary
The calendar items selector filters the time entries and calendar events from the state based on the selected date and transforms them into a `CalendarItem` that will be used by the collection view to layout and render the items.

### Technical considerations
The `CalendarLayoutCalculator` is a port from the old app, the only significant change is that now it returns a `CalendarItem` with it's calculated values instead of a different type.

### Relationships
Closes #202 

### Pull Request checklist

### Pointers
The code is a port from the old app, you can compare it's doing the same, the tests will guide you in the review.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


App/App.xcodeproj/xcshareddata/xcschemes/App.xcscheme
```diff
@@ -183,6 +183,16 @@
...
+         <TestableReference
+            skipped = "NO">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "8CFDB349247C072600952533"
+               BuildableName = "CalendarTests.xctest"
+               BlueprintName = "CalendarTests"
+               ReferencedContainer = "container:../Modules/Calendar/Calendar.xcodeproj">
+            </BuildableReference>
+         </TestableReference>
```
heytherewill, line 195: Who dis?

Modules/Calendar/Calendar/Common/CalendarItem.swift
```diff
@@ -0,0 +1,52 @@
...
+        }
+    }
+
+    let value: Value
+    let columnIndex: Int
+    let totalColumns: Int
+
+    var description: String { value.description }
+    var start: Date
+    var duration: TimeInterval?
```
heytherewill, line 47: Are these automagically implemented by `value` as well?

App/App.xcodeproj/xcshareddata/xcschemes/App.xcscheme
juanlaube: This is to also run the calendar tests when running all the tests

Modules/Calendar/Calendar/Common/CalendarItem.swift
juanlaube: Unfortunately no, because the duration can be different for UI purposes (e.g: < 15 min) so it's set when creating the item

Modules/Calendar/Calendar/CalendarDay/CalendarDaySelector.swift
```diff
@@ -1 +1,32 @@
...
+            return timeEntry.start >= date && stop <= date.addingTimeInterval(.secondsInADay)
+        } else {
+            let duration = time.now().timeIntervalSince(timeEntry.start)
+            return timeEntry.start >= date && duration <= .secondsInADay
+        }
+    }
+
+    let isCalendarEventIncluded: (CalendarEvent) -> Bool = { calendarEvent in
+        return calendarEvent.start >= date && calendarEvent.stop <= date.addingTimeInterval(.secondsInADay)
+
```
Odrakir, line 19: :fire:

Modules/Calendar/Calendar/CalendarDay/CalendarDaySelector.swift
```diff
@@ -1 +1,32 @@
...
+    }
+
+    let timeEntries = state.timeEntries.values
+        .filter(isTimeEntryIncluded)
+        .map(CalendarItem.Value.timeEntry)
+    let calendarEvents = state.calendarEvents.values
```
Odrakir, line 25: Maybe separate these two lines?

Modules/Calendar/Calendar/Common/CalendarItem.swift
Odrakir: This is not clear to me either. Are the properties in the root level for display only? Shouldn't start at least be the same? 🤔

Modules/Calendar/Calendar/CalendarDay/CalendarLayoutCalculator.swift
```diff
@@ -0,0 +1,126 @@
...
+
+    private let offsetFromNow: TimeInterval = 7 * .secondsInAMinute
+    private let minimumDurationForUIPurposes: TimeInterval = 15 * .secondsInAMinute
+    private let time: Time
+
+    init(time: Time) {
+        self.time = time
+    }
+
+    func calculateLayoutAttributesforItems(calendarItems: [CalendarItem.Value]) -> [CalendarItem] {
```
Odrakir, line 17: I guess this is how it was done in the old app and we are not changing things right now, but I think it's confusing that this method is called `calculateLayoutAttributesforItems` and it's mapping from `[CalendarItem.Value]` to `[CalendarItem]`. It's a naming issue more than anything else.

Modules/Calendar/Calendar/CalendarDay/CalendarLayoutCalculator.swift
```diff
@@ -0,0 +1,126 @@
...
+        let actualDuration = duration(for: item)
+        return actualDuration < minimumDurationForUIPurposes ? minimumDurationForUIPurposes : actualDuration
+    }
+
+    private func attributes(for value: CalendarItem.Value, columnIndex: Int, totalColumns: Int) -> CalendarItem {
+        return CalendarItem(
+            value: value,
+            columnIndex: columnIndex,
+            totalColumns: totalColumns,
+            start: value.start,
```
Odrakir, line 122: So yeah, regarding my other comment, `start` can be removed from the root `CalendarItem` level and from the initialiser, can't it?

Modules/Calendar/Calendar/Common/CalendarItem.swift
juanlaube: Nice catch, start can be derived from the wrapped value.

Modules/Calendar/Calendar/CalendarDay/CalendarLayoutCalculator.swift
juanlaube: I have been struggling to find a better name and couldn't find anything, we can always change it later.

PR #307, opened 2020-05-28
Add the StartTimeDragged action

### LINKED ISSUE #206: Add the StartTimeDragged action
## Description

Add the StartTimeDragged action and handle it in the reducer. This action should update the selected item with the new start time.

## Definition of done
- [x] The start time in the selected item is updated when it's a time entry.
- [x] Tests are created

sofokli1:
### Summary
- Updates the cal state to add the first property needed for the action/reducer
- Adds the StartTimeDragged action
- Adds the Either enum to handle multiple types in one property
- Adds the ArchitectureTestSupport Framework in order to maintain the shared reducer assertion code
- Adds build config files for the new framework
- Adds tests for the new action

### Relationships
Closes #206

### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


Modules/Utils/Utils/Either/Either.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
+
+public enum Either<A: Equatable, B: Equatable>: Equatable {
```
juanlaube, line 3: Why the `Equatable` constraints? 🤔

sofokli1: for the A and B ones 
Screenshot 2020-05-28 at 6 03 55 PM

for the enum level one 
Screenshot 2020-05-28 at 6 04 37 PM

juanlaube: oh I see, the problem is state requires `Equatable` things. I wouldn't constraint `Either` to that though, in that case I would remove the constraints from the definition and add this extension:

```
extension Either: Equatable where A: Equatable, B: Equatable { }
```

PR #321, opened 2020-05-29
🚧 Add the calendarViewAppeared and calendarEventsFetched actions

### LINKED ISSUE #203: Add the CalendarViewAppeared action
## Description

Add the CalendarViewAppeared action and handle it in the reducer. This action should create an effect to fetch the visible calendar items from the calendar service.

## Definition of done
- [ ] The `FetchCalendarEventsEffect` is dispatched and handled
- [ ] The state is updated with the fetched calendar events
- [ ] Tests are created
Make sure it's tested

Depends on #201

juanlaube:
## Info
This adds the effect to load the calendar events from the calendar service when the calendar view appears.

### Summary
- Adds the calendarViewAppeared action
- Adds the calendarEventsFecthed action
- Adds the fetchCalendarEventsEffect effect

### Technical considerations
It uses the calendar service to fetch the events from the available calendars.
I had to rename `CalendarService.Calendar` because of namespace conflicts with `Foundation.Calendar`, it's not `CalendarSource`.

### Relationships
Closes #203 

### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


Modules/Utils/Utils/Extensions/TimeInterval+Extensions.swift
```diff
@@ -11,6 +11,22 @@ public extension TimeInterval {
     static var secondsInAMinute: TimeInterval { TimeInterval(Int.secondsInAMinute) }
     static var secondsInAnHour: TimeInterval { TimeInterval(Int.secondsInAnHour) }
     static var secondsInADay: TimeInterval { TimeInterval(Int.secondsInADay) }
+
+    init(seconds: Int) {
```
dshr, line 15: Why do we need this, when we have the `TimeInterval.with(_:)` initialisers? Let's :fire: one of them!

Modules/Utils/Utils/Extensions/TimeInterval+Extensions.swift
```diff
@@ -11,6 +11,22 @@ public extension TimeInterval {
...
+
+    init(minutes: Int) {
+        self.init(seconds: minutes * .secondsInAMinute)
+    }
+
+    init(hours: Int) {
+        self.init(seconds: hours * .secondsInAnHour)
+    }
+
+    init(days: Int) {
```
dshr, line 27: This might not be the best approach for days — bc of daylight savings some days might not be 24 hours, and this way we're opening ourselves up for potenital bugs down the road.

Modules/Utils/Utils/Extensions/TimeInterval+Extensions.swift
juanlaube: I added this one for readability and consistency only, do you still want to 🔥?

Modules/Utils/Utils/Extensions/TimeInterval+Extensions.swift
juanlaube: Tbh, none of these should be used 😄

Modules/Utils/Utils/Extensions/TimeInterval+Extensions.swift
dshr: Well, let's delete the `with` ones, they make less sense than these

Modules/Utils/Utils/Extensions/TimeInterval+Extensions.swift
juanlaube: I 🔥 the extension, however the effect still returns events for 24 hours because that's how the UI is created

PR #338, opened 2020-06-01
🚧 Add calendar item cell

### LINKED ISSUE #288: Add the collection view item cell
## Description
Add the collection view item cell. This cell needs to be configured with a CalendarItem.
Don't implement the collection view in this PR, it will be done in a different issue.

## Dependencies
Depends on #202

## Definition of done
- [ ] The CalendarItemCell class is added
- [ ] The cell can be configured for a calendar event
- [ ] The cell can be configured for a non-selected time entry
- [ ] The cell can be configured for a selected time entry (shows the editing anchors)

## Designs
[Light mode](https://toggl.invisionapp.com/d/main#/console/19800511/414238725/preview)
[Dark mode](https://toggl.invisionapp.com/d/main#/console/19800511/418465584/preview)

juanlaube:
## Info
This adds the cell for the calendar items.

### Summary
It's almost a direct translation from the old app, the cell can configure itself for all cases.

### Relationships
Closes #288 

### Pull Request checklist

### Pointers
It's really difficult to try this now because we don't have a collection view in place, my recommendation is to compare this with the old app. We can adjust any thing that doesn't work later.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


Modules/Assets/Assets/Cells/CalendarItemCollectionViewCell.xib
```diff
@@ -0,0 +1,104 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
sofokli1, line 1: why the separate xib and not a prototype cell in the cal storyboard?

juanlaube: The layout is complex enough to have it's own file I think.

PR #372, opened 2020-06-04
Add the EditButtonTapped action

### LINKED ISSUE #214: Add the EditButtonTapped action
## Description

Add the EditButtonTapped action and handle it in the reducer.

## Definition of done
- [x] Move editableTimeEntry from Timer's local state to the global app state
- [x] Set the editableTimeEntry to the selectedItem's time entry in the state
- [x] Tests are created

sofokli1:
### Summary
Add the EditButtonTapped action. Moves the editableTimeEntry from TimerState into the AppState.

### Relationships
Closes #214

### Pull Request checklist

### Pointers
Just go commit by commit

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


Modules/Calendar/Calendar/ContextualMenu/ContextualMenuReducer.swift
```diff
@@ -15,6 +16,11 @@ func createContextualMenuReducer() -> Reducer<ContextualMenuState, ContextualMen
             state.selectedItem = nil
             return []
 
+        case .editButtonTapped:
+            guard case .left(let editableTimeEntry) = state.selectedItem else { return [] }
+            state.editableTimeEntry = editableTimeEntry
+            return []
```
juanlaube, line 22: This wasn't in the scope, but I just tested the behavior in the old app and we need to close the contextual menu when opening the edit view, please set the selected item to nil here too.

PR #375, opened 2020-06-04
Introduce the EntityStore

### LINKED ISSUE #369: Introduce EntityStore similar to IdentifiedArray
## Description
Create a new collection type called `EntityStore` to keep our state entities. Base the solution on [this](https://github.com/pointfreeco/swift-composable-architecture/blob/master/Sources/ComposableArchitecture/SwiftUI/IdentifiedArray.swift), but don't rely on the iOS 13+ `Identifiable` protocol, instead using the `Entity` protocol.

## Definition of done
- [ ] The `EntityStore` is added
- [ ] Tests are added
- [ ] All states and tests are updated to use the new type

dshr:
### Summary
This PR introduces the EntityStore type to our app. It's a riff on PointFree's [IdentifiedArray](https://github.com/peterkovacs/swift-composable-architecture/blob/master/Sources/ComposableArchitecture/SwiftUI/IdentifiedArray.swift), but specific for our Entity types.

### Relationships
Closes #369

### Pull Request checklist

### Pointers
Commit by commit will make most sense and not overwhelm everyone.

### Review template

## Merge permissions
Go for it


Modules/Models/Models/EntityStore.swift
```diff
@@ -0,0 +1,160 @@
+import Foundation
+
+public struct EntityStore<EntityType>: MutableCollection, RandomAccessCollection where EntityType: Entity {
```
juanlaube, line 3: I'm not a fan of the name, Store implies persisting something somewhere and it's not the case, perhaps `EntityCollection`? 🤔

Modules/Models/ModelsTests/EntityStoreTests.swift
```diff
@@ -0,0 +1,58 @@
...
+        assert(entityStore1 == entityStore2)
+    }
+    
+    func test_entityStoreSubscript_returnsCorrectEntity() {
+        let entities = [0, 1, 2, 3].map { TestEntity(id: $0, name: "Entity \($0)") }
+        let entityStore = EntityStore(entities)
+        assert(entityStore[2] == entities[2])
+    }
+    
+    func test_entityStoreIdSubscript_returnsCorrectEntity() {
```
juanlaube, line 31: We should also have some tests where id != position

Modules/Models/Models/EntityStore.swift
```diff
@@ -0,0 +1,160 @@
...
+        /// Direct access to an entity by its id.
+        ///
+        /// - Parameter id: The id of entity to access. Must be a valid id for an
+        ///   element of the array and will _not_ insert elements that are not already in the array, or
+        ///   remove elements when passed `nil`. Use `append` or `insert(_:at:)` to insert elements. Use
+        ///   `remove(id:)` to remove an element by its identifier.
+        /// - Returns: The entity.
+        public subscript(id id: Int64) -> EntityType? {
+            get { self.dictionary[id] }
+            set {
```
juanlaube, line 52: I'm not against crashing in debug with a useful error message, but set and _modify/yield have different semantics, could this lead to a different behavior in release? 🤔

Modules/Models/Models/EntityStore.swift
dshr: Good call, this makes more sense

Modules/Models/Models/EntityStore.swift
dshr: From what I understand the `DEBUG` behaviour is more restrictive because of the extra checks, but might be a bit slower because of said checks as well, so the PointFree folks only used them for debug/testing. I don't think we'll end up in a situation where our production code will be different from our debug code, causing different behaviour.

That said, I'm not against just sticking to the debug behaviour here.

Modules/Models/ModelsTests/EntityStoreTests.swift
dshr: I'll add those

PR #376, opened 2020-06-04
Add the calendar collection view

### LINKED ISSUE #289: Add the collection view to the calendar day view controller
## Description
Add the collection view to the calendar day view with its data source and delegate.
Don't implement the collection view layout in this PR, it will be done in a different issue, you can use the default flow layout instead.

## Dependencies
Depends on #202
Depends on #288 

## Definition of done
- [ ] The collection view is added
- [ ] The collection view data source reflects the data from the calendar items selector
- [ ] The item selection in the delegate dispatches the `CalendarDay.itemTapped` action

juanlaube:
## Info
Adds the collection view used for the calendar.

### Summary
This is only the collection view with an animatable rx datasource, the layout will be done in another PR.

### Relationships
Closes #289 

### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


Modules/Calendar/Calendar/CalendarDay/Views/CalendarDayCollectionViewDataSource.swift
```diff
@@ -0,0 +1,34 @@
+import UIKit
+import RxDataSources
+
+struct CalendarItemsSingleSection: AnimatableSectionModelType {
+
+    typealias Item = CalendarItem
+    typealias Identity = Int
```
sofokli1, line 7: where are these typealiases used? either use them or 🔥 them

PR #377, opened 2020-06-05
Fix time entry creation and updates

### LINKED ISSUE #364: Fix time entry creation and updates
## Description
Currently our StartEditReducer doesn't handle single time entry updates, only grouped ones. In the meantime, the TimeEntries reducer can't handle TEs that are created with manual mode or via calendar. This is blocking the calendar project, plus is a missing feature that slipped between the cracks.

## Definition of done
- [ ] Add TE update handling to the `TimeEntriesReducer`
- [ ] Amend the `StartEditReducer` to pass updates to the `TimeEntriesReducer`
- [ ] Rename the `StartTimeEntryDto` to `CreateTimeEntryDto` and use it for calendar TE creation

dshr:
### Summary
This PR adds a lot of missed TE creation/update functionality that got overseen during development.

### Relationships
Closes #364

### Pull Request checklist

### Pointers
Commit by commit should work best!

### Review template

## Merge permissions
Feel free


Modules/CommonFeatures/CommonFeatures/TimeEntries/TimeEntriesAction.swift
```diff
@@ -7,9 +7,12 @@ public enum TimeEntriesAction: Equatable {
     case continueTimeEntry(Int64)
     case startTimeEntry(StartTimeEntryDto)
     case createTimeEntry(CreateTimeEntryDto)
+    case updateTimeEntry(TimeEntry)
+    case updateTimeEntries([TimeEntry])
```
heytherewill, line 11: Why is the batch needed when you can always output multiple actions?

Modules/Timer/Timer/StartEdit/StartEditReducer.swift
```diff
@@ -176,10 +199,15 @@ func doneButtonTapped(_ state: StartEditState, _ repository: TimeLogRepository)
...
+            projectId: editableTimeEntry.projectId,
+            taskId: editableTimeEntry.taskId,
+            tagIds: editableTimeEntry.tagIds,
             billable: editableTimeEntry.billable) }
-
-    persistUpdatedTimeEntries(updatedTimeEnries)
-    return []
+    
+    return [
+        Effect.from(action: .timeEntries(.updateTimeEntries(updatedTimeEnries)))
```
heytherewill, line 209:
```suggestion
        Effect.from(action: .timeEntries(.updateTimeEntries(updatedTimeEntries)))
```

Modules/CommonFeatures/CommonFeatures/TimeEntries/TimeEntriesAction.swift
dshr: We do have batch repo/db update methods, so it only made sense to have a batch update action/effect. Also, multiple actions mean multiple state updates, which feels off to me when the actual underlying action is a batch update 🤷

heytherewill: We support effect batching, so there won't be multiple state updates

PR #378, opened 2020-06-05
Add logSuggestions boilerplate code

### LINKED ISSUE #329: Add suggestions boilerplate code
## Description
Add suggestions boilerplate code
## Dependencies
None
## Definition of done
- [x] Create empty feature with all required files
- [x] Hook everything to the main feature
- [x] Combine the [SuggestionProviderType](https://github.com/toggl/mobileapp/blob/develop/Toggl.Core/Suggestions/SuggestionProviderType.cs) and the [Suggestion](https://github.com/toggl/mobileapp/blob/develop/Toggl.Core/Suggestions/Suggestion.cs) Model into 1 enum with associated values like:
```swift
enum Suggestion {
    case mostUsed(description: String, color: String, ...etc) 
    case calendar(...) }
```
- [x] Inject the following dependencies in the reducer:
Calendar service (mock if not available)
Repository
~Analytics service (mock if not available)~
Time service

sofokli1:
### Summary
- Adds the boilerplate code
- Connects to the timer feature
- Adds the LogSuggestion model
- Adds dependencies to the reducer

### Relationships
Closes #329

### Pull Request checklist

### Pointers
Give special consideration to the LogSuggestion model, the rest is boilerplate.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


Modules/Timer/Timer/LogSuggestions/Model/LogSuggestion.swift
```diff
@@ -0,0 +1,55 @@
+import Foundation
+
+typealias SuggestionProperties = (
```
dshr, line 3: Why not a struct? It would handle conforming to `Equatable` much easier, and feels more appropriate for a structured storage of data.

sofokli1: was trying not to add an other model. but it kind of makes sense at this point

PR #448, opened 2020-06-16
Add a way to export an UIColor as HSB

### LINKED ISSUE #297: Implement an HSB constructor and HSB export to our colours
## 🎯 What needs to be done
Add extensions to `UIColor` that would allow it to be constructed from HSB values or be expressed as an HSB tuple/struct.

dshr:
## Info
This PR adds a way to express as UIColor as an HSB tuple.

### Relationships
Closes #297

### Pull Request checklist

### Pointers
Copy the extension to a new playground and test it against some known colours. Bear in mind that iOS scales HSB values from 0 to 1 and not 0 to 360.

### Review template

## Merge permissions
Go for it


dshr:
I'm also downgrading the original issue to XS

Modules/Utils/Utils/Extensions/UIColor+Extensions.swift
```diff
@@ -73,4 +73,15 @@ public extension UIColor {
         self.luminance < 0.5 ? .white : .black
     }
     // swiftlint:enable no_ui_colors
+    
+    // swiftlint:disable large_tuple
+    var hsb: (hue: CGFloat, saturation: CGFloat, brightness: CGFloat) {
```
juanlaube, line 78: Oh look! Another extension that should be part of the SDK 🙃

dshr: I'm adding a `.hex` String as well 🤷‍♂️

PR #450, opened 2020-06-16
Add the contextual menu UI

### LINKED ISSUE #295: Add the contextual menu bottom sheet
## Description
Add the contextual menu bottom sheet view controller.
The contextual menu should be visible when `selectedItem` in the state is not nil.
Don't worry about configuring the buttons yet, that will be done in a different issue.

## Dependencies
Depends on #289 

## Definition of done
- [ ] The contextual menu bottom sheet is added
- [ ] The visibility depends on the selected item

## Designs
[Light mode](https://toggl.invisionapp.com/d/main#/console/19800511/415483648/preview)
[Dark mode](https://toggl.invisionapp.com/d/main#/console/19800511/418465585/preview)

juanlaube:
## Info
This adds the contextual menu UI

### Summary
This adds the contextual menu UI and binds the visibility to the selectedItem in the state.

### Technical considerations
- This is just the layout and visibility
- The visibility depends on having a selected item
- When visible, it hides the tab bar so it looks like in the designs
- Some assets are not final yet, since I couldn't export the correct icons from invision, I left a comment requesting the correct icon
- Background color in the icons are not final, there's another issue to configure the contextual menu

I was going to use the bottom sheet controller, in fact, I managed to move it to utils, but the behavior is not the same, it adds the overlay in the background and ignores all touches to interact with the calendar.

### Relationships
Closes #295 

### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


dshr:
Is the cancel button supposed to be outside the safe area like that?
Screenshot 2020-06-17 at 19 50 07

juanlaube:
Yes, it will look correct in runtime, of course, check the designs here https://toggl.invisionapp.com/d/main#/console/19800511/420486611/inspect

dshr:
Screenshot 2020-06-18 at 16 32 09
Just re-tested in at runtime, it's def off. I think I opened the storyboard exactly because the cancel button looked awfully close to the bottom 🤷‍♂️

juanlaube:
@dshr fixed, the constraint's constant was correct, but the order was inverted 🤦‍♂️

dshr:
You're gonna hate me for this, but it still doesn't line up:
Screenshot 2020-06-19 at 11 20 21
Also, the title label seems cut off, something's clipping it

juanlaube:
@dshr check it now Screen Shot 2020-06-22 at 09 07 57

sofokli1:
if the description is empty shouldn't the project label go all the way up aligned with the duration label?

juanlaube:
Damn, you are right @sofokli1, I will fix it quickly

juanlaube:
@sofokli1 fixed, the project and the description are now in a stack view so it's easy to show/hide them.

PR #477, opened 2020-06-26
Create the combined Hue/Saturation control

### LINKED ISSUE #319: Create the combined Hue/Saturation control
## Description
Implement the HS control as a `UIControl` subclass and connect it to the state. Designs are here: https://toggl.invisionapp.com/d/main#/console/20014282/419193923/preview, while implementation specs can be found here: https://www.notion.so/toggl/Color-Picker-Unification-fe79a56649d74175a1b004936c7ebe10#1a1dc1f21c7f44f89adc95b666248458
TLDR: Hue changes from 0 to 360, and saturation changes for 30 to 100

## Dependencies
Depends on #318

## Definition of done
- [x] The HS control is implemented as a 2d control
- [x] The HS control parses the H and S values of the currently set custom colour and sets its thumb to the appropriate coordinates
- [x] The control's thumb is coloured into the currently set custom colour
- [x] The control's background changes depending on dark mode (see Notion doc)
- [x] The control's background gets lighter/darker depending on the currently set custom colour's V value
- [ ] Moving the control's thumb emits a .colorHueSaturationChanged action with a tuple containing the new Hue and Saturation pair. This action then emits a `.colorPicked` effect.

sofokli1:
### Summary
Creates the view and hooks it to the appropriate action in the VC

Opted for drawing the whole view instead of using layer. Had some troubles with layers and this was easier.

### Relationships
Closes #319

### Pull Request checklist

### Pointers
For testing the `value` changes you need to manually set it in code and run it again (until the slider is implemented)

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


Modules/Utils/Utils/Extensions/Comparable+Extention.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+public extension Comparable {
+    func clamp(min: Self, max: Self) -> Self {
```
juanlaube, line 4: Amazing! 😄

Modules/Timer/Timer/Project/Views/HueSaturationPickerView.swift
```diff
@@ -0,0 +1,154 @@
...
+class HueSaturationPickerView: UIControl {
+    // swiftlint:disable no_ui_colors
+    private let circleDiameter: CGFloat = 30
+    private var circleRadius: CGFloat { circleDiameter / 2 }
+
+    private let outerCircleDiameter: CGFloat = 32
+    private var outerCircleRadius: CGFloat { outerCircleDiameter / 2 }
+    private let outerCircleColor = UIColor.white.projectColor
+
+    private static let defautlColor = UIColor(hex: "3178be").projectColor
```
juanlaube, line 13:
```suggestion
    private static let defaultColor = UIColor(hex: "3178be").projectColor
```

sofokli1:
@juanlaube the same thing happens in the old app. @dshr can you clarify how it should behave?

Modules/Timer/Timer/Project/Views/HueSaturationPickerView.swift
```diff
@@ -0,0 +1,154 @@
...
+        drawOuterCircleWithShadow(rect: outerCircleRect)
+
+        let circleRect = CGRect(x: x, y: y, width: circleDiameter, height: circleDiameter)
+        drawColorCircle(rect: circleRect)
+    }
+
+    private func drawRainbowBackground(rect: CGRect) {
+        let context = UIGraphicsGetCurrentContext()!
+
+        let red = UIColor.red.projectColor
```
dshr, line 84: For some reason these don't change when switching between light/dark mode.

Modules/Timer/Timer/Project/Views/HueSaturationPickerView.swift
```diff
@@ -81,13 +81,13 @@ class HueSaturationPickerView: UIControl {
...
         let context = UIGraphicsGetCurrentContext()!
 
-        let red = UIColor.red.projectColor
-        let yellow = UIColor.yellow.projectColor
-        let green = UIColor.green.projectColor
-        let cyan = UIColor.cyan.projectColor
-        let blue = UIColor.blue.projectColor
-        let magenta = UIColor.magenta.projectColor
-        let borderColor = UIColor(hex: "CECECE").projectColor
+        let red = UIColor.red
```
dshr, line 101: The gradient colours should have the dark mode transformations applied to them. My previous comment meant that the code looked okay, and that the transforms should've worked, but they didn't for some reason, and I guess I should've phrased it better, sorry about that.

PR #525, opened 2020-06-30
Fix some app inconsistencies

### LINKED ISSUE #523: Catch up on app translations
## Description
We need to catch up on translations and stop using hardcoded strings

## Definition of done
- [ ] All hardcoded strings are removed from the app
- [ ] These strings are moved to the resources and properly translated

### LINKED ISSUE #524: Stop including running time entries in the log
## Description
The running time entry should not be included in the log, ever.

## Definition of done
- [ ] The running time entry is not included in the main log

heytherewill:
### Summary
This fixes some inconsistencies in the codebase, namely:

- It removes the running entry from the main log
- It extracts all hardcoded strings, putting them into a localizable file
- It removes the iOS13 icons, to remove warnings 
- It changes the layout of the main log so that the description is centered when there's no project
- It changes the layout of the suggestions so that the description is centered when there's no project
- It changes the assets for the running entry card to use proper buttons

### Relationships
Closes #523
Closes #524

### Pull Request checklist

### Pointers
Go for it commit by commit. Do make a functional test.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone, anytime


Modules/Assets/Assets/Extensions/Strings.swift
```diff
@@ -7,10 +7,21 @@ public struct Strings {
...
+    public static func create(project: String) -> String {
+        return "\("createProject".localized) \(project)"
+    }
+    
+    public static func create(tag: String) -> String {
+        return "\("createTag".localized) \(tag)"
+    }
+    
+    public static var today: String { "yesterday".localized }
+    public static var yesterday: String { "today".localized }
```
sofokli1, line 19: these 2 need swapping

sofokli1, line 19:
```suggestion
    public static var today: String { "today".localized }
    public static var yesterday: String { "yesterday".localized }
```

Modules/Timer/Timer/TimeEntriesLog/Views/TimeEntryCell.swift
```diff
@@ -22,4 +23,9 @@ class TimeEntryCell: UITableViewCell {
 
         // Configure the view for the selected state
     }
+    
+    override func layoutSubviews() {
+        super.layoutSubviews()
+        _ = containerView.roundedCorners()
```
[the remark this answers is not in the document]
Odrakir: @dshr clarified to me the intention of this was to make those methods chainable. Maybe we can then add @discardableResult to it so we don't have to do the `_ = ...`

PR #526, opened 2020-07-01
Create the Value control

### LINKED ISSUE #320: Create the Value control
## Description
Implement the Value control as a `UISlider` subclass and connect it to the state. Designs are here: https://toggl.invisionapp.com/d/main#/console/20014282/419193923/preview, while implementation specs can be found here: https://www.notion.so/toggl/Color-Picker-Unification-fe79a56649d74175a1b004936c7ebe10#1a1dc1f21c7f44f89adc95b666248458
TLDR: Values go from 30 to 100

## Dependencies
Depends on #318

## Definition of done
- [x] The Value control is implemented as a `UISlider` subclass
- [x] The Value control parses the V values of the currently set custom colour and sets its thumb to the appropriate coordinates
- [x] The control's max value tint gets the HS value of the currently set custom colour
- [ ] Moving the control's thumb emits a `.colorValueChanged` action with the new V value. This action then emits a `.colorPicked` effect.

sofokli1:
### Summary
Adds the Value slider and hooks it to the rest of the custom color picker.
### Relationships
Closes #320

### Pull Request checklist

### Pointers
Test on dark mode and light mode

### Review template
- [ ] The code considers i18n

## Merge permissions
🦑


Modules/Timer/Timer/Project/Views/ValueSlider.swift
```diff
@@ -0,0 +1,110 @@
...
+        drawGradientBackground(rect: bounds)
+    }
+    
+    // swiftlint:disable no_ui_colors
+    func drawGradientBackground(rect: CGRect) {
+        let context = UIGraphicsGetCurrentContext()!
+        // This non-generic function dramatically improves compilation times of complex expressions.
+        func fastFloor(_ val: CGFloat) -> CGFloat { return floor(val) }
+
+        let black = UIColor.black.projectColor
```
dshr, line 81: The right-hand side of the slider shouldn't be black, but the target colour with the min possible brightness, that being 0.3 in our case

Modules/Timer/Timer/Project/Views/ValueSlider.swift
```diff
@@ -0,0 +1,110 @@
...
+        drawThumb(rect: CGRect(x: 0, y: 0, width: imageSize.width, height: imageSize.height))
+
+        let image = UIGraphicsGetImageFromCurrentImageContext()!
+        UIGraphicsEndImageContext()
+
+        return image
+    }()
+
+    private var selectedColor: UIColor {
+        UIColor(hue: hue, saturation: saturation, brightness: complement(CGFloat(value)), alpha: 1).projectColor
```
dshr, line 41: Applying to project colour transform here seems wrong to me. Our designs show the V control's right side as always darker than the left, while this will make it lighter on dark mode. I'm going to reach out to the design team about this, the specs aren't as clear as I initially thought they were. I'm going to ping you in the query as well, just to keep you in the ➰.

Modules/Timer/Timer/Project/Views/ValueSlider.swift
```diff
@@ -0,0 +1,110 @@
...
+        context.drawLinearGradient(gradient,
+            start: CGPoint(x: rectangleRect.maxX, y: rectangleRect.midY),
+            end: CGPoint(x: rectangleRect.minX, y: rectangleRect.midY),
+            options: [])
+        context.restoreGState()
+    }
+
+    func drawThumb(rect: CGRect) {
+        let outerThumbColor = UIColor(hex: "CECECE")
+        let innerThumbColor = UIColor.white
```
dshr, line 99: Let's add these colours to our colour assets catalogue.

Modules/Timer/Timer/Project/Views/ValueSlider.swift
```diff
@@ -0,0 +1,110 @@
+import UIKit
+
+class ValueSlider: UISlider {
+
+    private static let defaultColor = UIColor(hex: "3178be").projectColor
```
dshr, line 5: I guess the default custom colour should also be a named colour, just so we keep it in one place.

Modules/Timer/Timer/Project/Views/ValueSlider.swift
Odrakir: Yep, I completely agree with this point. It looks off in dark mode.

Modules/Timer/Timer/Project/Views/ValueSlider.swift
Odrakir: I'm sorry because I missed this conversation, but why are we limiting this to 30% ?

dshr: Because of the specs: https://www.notion.so/toggl/Color-Picker-Unification-fe79a56649d74175a1b004936c7ebe10#fdb5964548be40dfa50b26e452bcca83

Modules/Timer/Timer/Project/Views/ValueSlider.swift
```diff
@@ -25,41 +28,54 @@ class ValueSlider: UISlider {
...
+    }
+
+    func setup() {
+        value = complement(ValueSlider.defaultColor.hsb.brightness)
+
+        tapGesture.addTarget(self, action: #selector(self.handleTap(_:)))
+        tapGesture.delegate = self
+        self.addGestureRecognizer(tapGesture)
+
+        panGesture.addTarget(self, action: #selector(self.handleTap(_:)))
```
Odrakir, line 57: This one is also calling `handleTap`. `handlePan` is not used.

sofokli1: Addressed

sofokli1:
@Odrakir and @dshr since product will get back on this Tomorrow. I say we merge this and make an new issue if there need to be changes

PR #532, opened 2020-07-02
Add SettingTapped action

### LINKED ISSUE #419: Add SettingTapped action
## Description
Add AppAction.SettingTapped action. Includes the Setting enum. Possible values:
- Workspace
- DateFormat
- DurationFormat
- FirstDayOfTheWeek
- CalendarSettings
- SmartAlerts
- SubmitFeedback
- About
- PrivacyPolicy
- TermsOfService
- Licenses
- Help

I leave this app to the dev working on this. But my general idea was to use the same pattern we have for editableTimeEntry until we migrate to a new way for routing. So have this action set the `EditableUserPreferences` property to the be same as `UserPreferences`. This will trigger the settings UI. Have all actions perform changes to the `EditableUserPreferences` and when closing the view, `UserPreferences` gets replaced with `EditableUserPreferences`.

## Dependencies
#417
## Definition of done
- [x] Add action
- [x] Add enum
- [ ] Add tests

sofokli1:
### Summary
~Adds the action and the SettingItem enum~
Action is not needed anymore. Adds some a protocol and some structs to structure the settings view later

### Relationships
Closes #419 since it is not needed anymore

### Pull Request checklist

### Pointers
Just take a look at the code

### Review template
- [ ] The code considers i18n

## Merge permissions
🦑


sofokli1:
@toggl/ios-senseis please take a look at this draft and let me know what you think.
Main focus is the enum for the `SettingItem`. My thinking is to have this enum that we can use it to get all the high level settings list and their names for the UI. Also add an optional `SettingItem` in the state to keep track of which setting we are showing. Based on the value the Setting VC will decide which sub-setting VC to push.  **How does this sound?**

This is just the start. I was thinking of adding later on a `SettingSection` enum looking something like the following, to be used to organise and populate the settings tableview.
```swift
enum SettingSection {
    case plan([SettingItem])
    case yourProfile([SettingItem])
    ...

    var name: String {
    // here goes a switch to return the section header name localized
    }
}
```

and as this goes add enums for each setting that had 1 of multiple choices (like date format, duration format etc.) to be used to populate the selection view for that specific setting.

basically adding and removing things from the settings view will be super easy, just by updating enums and the tree of `SettingItem`s nested in `SettingSection`s that will be in the the VC.

Does this sound reasonable?
Do you have a better idea?

Odrakir:
I'm not sure if this is the correct approach. What good do those enums have if in the end you are going to need an array to sort all of the possibilities? You'll need something like:

´´´swift
let tableSections: [SettingSection] = [
    .yourProfile([
        .userName,
        .email,
        .workspace
    ]),
    .dateAndTime([
       .dateFormat,
       .use24HourClock
      ....
    ])
]
´´´

This makes the enum not that useful, because it doesn't guarantee we are not missing some of the options when creating that array. Also we need to specific what kind of field we are dealing with somewhere, probably inside of the enum, plus also the value for each of them.

I think I would rather use an enum for the type of field (navigation, toggl, label, button) and have a struct to hold the rest of the stuff the field needs.

```swift
enum SettingsItemType {
   case label(String)
   case navigation(String)
   case toggle(Bool)
}

struct SettingsItem {
    var type: SettingsItemType
    var title: String
    var action: () -> ()
}

// Then you'll generate the settings table view with whatever comes from the store
let tableSections = [
    SettingsSection("Profile", store.whateverProperyWeNeed.mapToFields())
   ....
]
```

Alterntatively to the enum we could use a protocol like we used to in the old app:  https://github.com/toggl/mobileapp/tree/617bdfa66cd3ca36b9d141c42bb2cf712ccf4e42/Toggl.Core.UI/Views/Settings

This is how the settings were generated in the old app: https://github.com/toggl/mobileapp/blob/986d5e0dd6165e04b675bf9094d725d974b28267/Toggl.iOS/ViewControllers/Settings/SettingsViewController.cs#L82

dshr:
I'm with @Odrakir on this one — the section enum seems pointless, since the only difference between sections is the section title, which means we can get away with an array of structs for that. 

As for the cell item, I also agree that we can go more generic, and not create an enum case for each cell, but we can still use an enum with some named associated values:
``` Swift
enum SettingsItem {
   case label(String)
   case navigation(title: String, action: () -> ())
   case toggle(title: String, isOn: Bool, action: () -> ())
}
```

sofokli1:
@Odrakir do you agree with @des's approach for not using a struct but keeping the struct fileds as associated values in the enum?

Odrakir:
There are some common fields (label, action), that's why I was proposing a struct with an enum inside for the not common fields. Or, as I said, protocol inheritance, as in the old app might be a simpler solution.

dshr:
We might have fields with no title further down the line, and even now not all fields have an action, so they'd have to be optionals  🤷🏻‍♂️ 
It's all the same in the end, main difference would be when dequeuing stuff:
``` Swift
switch item:
case let .label(title):
    let cell = .... // deq the cell here
    cell.titleLabel.text = title
    return cell
case let .navigation(title: title, action: action):
    let cell = .... // deq the cell here
    cell.titleLabel.text = title
    cell.action = action
    return cell
.........
```
IMO the enum with wrapped values ends up being cleaner
vs
``` Swift
switch item.type {
case .label:
    let cell = .... // deq the cell here
    cell.titleLabel.text = item.title
    return cell
case let .navigation(route):
    let cell = .... // deq the cell here
    cell.titleLabel.text = item.title
    if let action = item.action {
        cell.action = action
    }
    cell.route = route // just spitballing here
    return cell
........
```

Odrakir:
I still prefer to have just the type as an enum rather than having different associated values, which could be too many soon, but I'm not going to fight this. Eugene, you choose.

dshr:
When we get too many associated values, we can just wrap them into a struct? Like `case cellType(CellTypeData)`?

sofokli1:
I am going with something similar to the old app approach since the point is not to try new approaches at this time but to get the app ported as fast as possible. we can revisit this in the future.

sofokli1:
@Odrakir and @des I removed the action altogether since it is not needed to keep track of the selected item in the menu. The new model ported from the old app have an action closure to execute the needed pushes or dispatch action etc. Let me know if this looks good. Will update the docs by removing the action and closing the issue with this PR if you agree.

Odrakir:
I think having an action so this still goes through the reducer is a good idea. You could use the `action: ()-> ()` to send that action to the store, or maybe even replace it by the specific `SettingsAction` that needs to be sent.

sofokli1:
@Odrakir that sounds good but with closures we can also do calls to push a VC etc. which is not a responsibility of the `SettingsAction`

Odrakir:
It depends. We are already doing some navigation through reducers and the router. The good thing is that they'll go through the architecture and we'll be easier to test and debug.

sofokli1:
@Odrakir in that case it makes sense to use `SettingsAction`. Updated

sofokli1:
@dshr when possible take a look as well

PR #541, opened 2020-07-06
Add SmartAlertsOpenSettingsAppTapped action

### LINKED ISSUE #434: Add SmartAlertsOpenSettingsAppTapped action
## Description
Add `SmartAlertsOpenSettingsAppTapped` action. It should take the user to the settings app as the old app (or whatever the platform usually does in this case)

## Dependencies
#417
## Definition of done
- [x] Add SmartAlertsOpenSettingsAppTapped action
- [x] Add tests

sofokli1:
### Summary
- Adds the action from the issue
- Connects settings to app level
- Adds routes for settings (only what is needed for the action from the issue)

### Relationships
Closes #434

### Pull Request checklist

### Pointers
Just check the code (trying to replicate what we do in onboarding and other parts of the app regarding routing). It is not much code added and most is boilerplate.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


Modules/Architecture/Architecture/App/AppRoutes.swift
```diff
@@ -29,3 +30,9 @@ public enum TabBarRoute: String, Route {
     case reports
     case calendar
 }
+
+public enum SettingsRoute: String, Route {
+    public var root: Route? { AppRoute.settings }
+
+    case osAppSettings
```
Odrakir, line 37: Can we call this `systemSettings`? It took me a while to understand what `osAppSettings` was.

PR #549, opened 2020-07-07
Set local dummy data

### LINKED ISSUE #530: Use workspace generator to add dummy local data
## Description
Since sync is not working yet I think it might come useful to have a bunch of local data to use for developing and debugging the UI. We can use the workspace generator for this and basically replace the text files in Assets/textfiles with whatever it generates.

Then we'll need to touch a bit the Repository so we can easily swap this mock version with the live one. (Repository.live and Repository.mock like the Pointfree guys do seems reasonable enough)

This will come handy not only on this early stages of development but also when ever we want to debug with specific data or when we don't have an internet connection.

## Definition of done
- [ ] The files in Assets/textfiles are replaced with the info generated by the workspace generator
- [ ] This mock version is easily swappable with the live one

Odrakir:
## Info
Sets data extracted from the Workspace generator as local JSON files to test the app.

### Summary
This just replaces the content of the JSON files in the assets with data from the workspace generator and then makes some changes in the repository so that data is fetched on fresh installs.

If you want to use this data just delete the app and install it again. If the DB is empty it'll get the data from those dummy files. Feel free to modify that data or create new copies to test different things.

### Relationships
Closes #530 

### Pull Request checklist

### Review template
As mentioned above, delete the app and install again for it to use the dummy data.

- [ ] The code considers i18n

## Merge permissions
Go!


Modules/Repository/Repository/Repository.swift
```diff
@@ -50,26 +42,92 @@ extension Repository: TimeLogRepository {
...
+
+                return Single.just(workspaces)
+            }
     }
     
     public func getClients() -> Single<[Client]> {
         database.clients.rx.getAll()
+            .flatMap { clients in
+                if clients.count == 0 {
+                    return self.api.loadClients()
```
heytherewill, line 62: I really don't like the inclusion of an API here. It's misleading, since this api will never be called with actual data. Why not mock the entire repository instead?

Odrakir: Take another look, I subclassed the Repository. Also I've called this version of the environment `local` rather than `mock`, since I think it better reflects how it works: the database and time services are the live ones, but the API takes the data from local files. We could further create other versions of the environment for different purposes.

PR #551, opened 2020-07-07
Fix TimeEntryViewModel comparison

### LINKED ISSUE #550: Fix TimeEntryViewModel comparison
## Description

The comparison for equality is wrong, mainly because `projectTaskClient` is a reference type and can't be compared directly and also because it doesn't make much sense to compare end time and duration, both.

This causes the table view to redraw with every change to the state, and that has bad performance implications.

## Definition of done
- [ ] The comparison is fixed

Odrakir:
## Info
Fixed that comparison. Everything is described in the issue.

### Relationships
Closes #550 

### Pull Request checklist

### Pointers
If you really want to test the effect of this, have a bunch of TEs (more that 500) and see how when showing or hiding the start/edit card it takes 2 seconds or more to do it. Fix the comparison and try again.

### Review template

- [ ] The code considers i18n

## Merge permissions
Go!


Odrakir:
Nope, as described in the issue this is more because of the `projectTaskClient`. Adding `.string` did the trick. The others are extras.

PR #555, opened 2020-07-07
Style autocomplete suggestions part 1 (floating view and TE suggestions)

### LINKED ISSUE #443: Create the Time Entry autocomplete cell
## Description
Create a reusable cell to be used for time entry autocompletion. Designs are here: https://toggl.invisionapp.com/d/main#/console/19800511/416531012/preview

## Definition of done
- [ ] The cell has a description label
- [ ] The cell has a project colour view
- [ ] The cell has a project title label
- [ ] The spacing between the description label, colour view and project title label is always the same (this means that the colour views shouldn't line up)
- [ ] The cell is used when appropriate

### LINKED ISSUE #445: Style the suggestions list container
## Description
Add the proper styling to the suggestions list container. See here: https://toggl.invisionapp.com/d/main#/console/19800511/416531014/preview.

## Definition of done
- [ ] Container has the right spacing from the edges of the screen
- [ ] Corners are rounded
- [ ] Shadow is added

Odrakir:
## Info
This is the first part of the work to style the autocomplete suggestions, all in small chunks so you don't have a hard time reviewing.

### Summary
It's mostly UI work, so there's not much to say. Test it. This is only time entry suggestions for now, the rest will be built on top of this.

I've also updated the shadow color and all shadows since they were not following designs. I made the color itself 100% opaque and put the appropriate opacity in the shadow layer itself, as that changes from shadow to shadow.

### Relationships
Closes #445
Closes #443

### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
Go!


Modules/Timer/Timer/StartEdit/Effects/UpdateAutocompleteSuggestionsEffect.swift
```diff
@@ -34,14 +34,14 @@ func updateAutocompleteSuggestionsEffect(
 func fetchProjectSuggestions(for query: String, in entities: TimeLogEntities) -> [AutocompleteSuggestion] {
     func clientNameMatches(_ word: String, _ project: Project) -> Bool {
         guard let client = entities.getClient(project.clientId) else { return false }
-        return client.name.contains(word)
+        return client.name.lowercased().contains(word.lowercased())
```
heytherewill, line 37: There's no way to do case independent string comparisons without allocating two new strings?

Modules/Models/Models/Extras/AutocompleteSuggestion.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 
 public enum AutocompleteSuggestion: Equatable {
-    case timeEntrySuggestion(timeEntry: TimeEntry)
+    case timeEntrySuggestion(query: String, timeEntry: TimeEntry, project: Project?)
```
heytherewill, line 4: To me it seems like you are polluting the models with information needed for display. We can fetch the project from `timeEntry.projectId` and the query is part of the state. This info should be added by the selector and not by the query effect

Modules/Timer/Timer/Common/Views/BottomSheetViewController.swift
```diff
@@ -75,9 +75,9 @@ class BottomSheetViewController: UIViewController, UIGestureRecognizerDelegate {
         containedViewController.view.layer.cornerRadius = 10
         containedViewController.view.clipsToBounds = true
         containedViewController.view.layer.shadowColor = Color.shadow.cgColor
-        containedViewController.view.layer.shadowOffset = CGSize(width: 0, height: -6)
+        containedViewController.view.layer.shadowOffset = CGSize(width: 0, height: -2)
```
heytherewill, line 78: Magic Constants

Modules/Timer/Timer/StartEdit/Effects/UpdateAutocompleteSuggestionsEffect.swift
heytherewill: If there's no way, at least lowercase the word outside the function that gets called once for every entity

Odrakir: Oh! I was going to say that the only way to do it is using `range(of:)` instead of `contains, but apparently there's a way to do it. `localizedCaseInsensitiveContains`

Modules/Models/Models/Extras/AutocompleteSuggestion.swift
Odrakir: That's tricky... the state is an observable, we don't have a way to access the current state directly. Maybe we should and then this issue would be a bit easier. Alternatively `AutocompleteSuggestion` should have the actual model dependencies it needs to render itself, the way we do with `TimeEntryViewModel` (or the `TimeEntry` there could be a `TimeEntryViewModel` instead)... That's close to what I was doing but instead of adding another struct I did add it to the action.

I see your point though, let me try find a workaround.

Modules/Timer/Timer/Common/Views/BottomSheetViewController.swift
Odrakir: That's the offset of the shadow according to designs, it's not constant for every screen so I don't see the point of extracting it into a property.

Modules/Timer/Timer/AutocompleteSuggestions/Views/AutocompleteHeaderCell.swift
```diff
@@ -11,10 +11,22 @@ class AutocompleteHeaderCell: BaseTableViewCell<String> {
...
+            closeButton.image(for: .normal)?.withRenderingMode(.alwaysTemplate),
+            for: .normal
+        )
+
+        insertSeparator()
     }
 
     override func configure(with model: String) {
-        label.text = model
+        label.attributedText = NSAttributedString(
```
dshr, line 27: I think we can drop the `.kern` thing — all text styles are outlined [in the DSM](https://toggl.invisionapp.com/dsm/toggl/toggl-i-os/nav/foundations/folder/typeStyles/5ecb8dd2d605b3c381c55c2b?mode=preview) and none have `letter-spacing` set, so we can treat the designs as mistaken on this.

Modules/Timer/Timer/AutocompleteSuggestions/Views/AutocompleteSuggestionCell.swift
```diff
@@ -37,39 +32,27 @@ class AutocompleteSuggestionCell: BaseTableViewCell<AutocompleteSuggestion> {
...
+
             leftText.attributedText = attributedText
+
             if let project = project {
-                rightText.text = project.name
-                separator.backgroundColor = UIColor(hex: project.color)
-            } else {
-                rightText.isHidden = true
-                separator.isHidden = true
+                rightText.attributedText = projectClientTaskString(projectName: project.name, projectColor: project.color, spacerBefore: true)
```
dshr, line 39: Why not just have one label in here and join the two NSAttributedStrings?

Modules/Timer/Timer/AutocompleteSuggestions/Views/AutocompleteHeaderCell.swift
```diff
@@ -11,10 +11,22 @@ class AutocompleteHeaderCell: BaseTableViewCell<String> {
     
     override func awakeFromNib() {
         super.awakeFromNib()
-        backgroundColor = Color.autocompleteBackground.uiColor
+
+        contentView.backgroundColor = Color.autocompleteBackground.uiColor
+
+        closeButton.tintColor = Color.textPrimary.uiColor
+        closeButton.setImage(
```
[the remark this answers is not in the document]
Odrakir: As far as I know I can't set the image to be template and use the tint color in Storyboards.

Modules/Timer/Timer/AutocompleteSuggestions/Views/AutocompleteHeaderCell.swift
Odrakir: But then the text looks larger, as you pointed out in the previous review. I say we leave it and ask design about it. I'll take a note.

Modules/Timer/Timer/AutocompleteSuggestions/Views/AutocompleteSuggestionCell.swift
Odrakir: Because this allows me to better play with the size of each label. This is something design should pitch in as well, but for now I force the project to be at least 60 points if it exists, otherwise with long descriptions you wouldn't see the project at all.

dshr: Fair point, this makes sense

Modules/Timer/Timer/AutocompleteSuggestions/Views/AutocompleteHeaderCell.swift
dshr: Yeah, 🔥 the `.kern` thing and then post in #design-discussions

Modules/Timer/Timer/AutocompleteSuggestions/Views/AutocompleteHeaderCell.swift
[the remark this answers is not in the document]
dshr: You actually can, it's just a bit counterintuitive! You have to set it in the `Images.xcassets` file:
Screenshot 2020-07-09 at 18 17 52
As a rule, all our icons should be template images. I'll research if we can set the linter to check for that, and have two files like `Icons.xcassets` and `Images.xcassets`, so we can still have non-template images when needed.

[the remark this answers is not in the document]
dshr: Same thing with the `contentView.backgroundColor`, it can be set in the Storyboard too.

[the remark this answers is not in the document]
Odrakir: Oh! Good one.

Issue #453, opened 2020-06-17
Fetch user calendars


juanlaube:
## Description

We need to complete the `getAvailableCalendars()` method. This should access the event store and fetch all the available calendars. 

## Dependencies
Depends on #452 

## Definition of done
- [x] If permissions are granted, fetch all the existing `EKCalendar` from an `EKStore` and transform them into `CalendarSource`
- [x] If permissions are not granted, return an empty array

PR #563, opened 2020-07-08
Implement cal service functions

### LINKED ISSUE #454: Fetch calendar events
## Description

We need to complete the `getCalendarEvents()` method. This should fetch all the calendar events between the given dates that are in the selected calendars.
The method needs to filter all events between the given start and end dates, that are not all day events.

## Dependencies
Depends on #453 

## Definition of done
- [ ] If permissions are granted, fetch all the existing `EKEvent` from an `EKStore` that match the predicate and transform them into `CalendarEvent`
- [ ] If permissions are not granted, return an empty array

sofokli1:
### Summary
Mostly ported the logic from the old app

### Relationships
Closes #454 #453 

### Pull Request checklist

### Pointers
Quick look at the code should do the trick.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


Modules/CalendarService/CalendarService/CalendarService.swift
```diff
@@ -11,7 +12,7 @@ public protocol CalendarService {
 
 public class DefaultCalendarService: CalendarService {
 
-    var eventStore = EKEventStore()
+    private var eventStore = EKEventStore()
```
heytherewill, line 15: Could this not be a let binding?

Modules/CalendarService/CalendarService/CalendarService.swift
```diff
@@ -34,10 +35,55 @@ public class DefaultCalendarService: CalendarService {
...
+
+    private func toCalendarEventFromEvent(event: EKEvent) -> CalendarEvent {
+        CalendarEvent(
+            id: event.eventIdentifier,
+            calendarId: event.calendar.calendarIdentifier,
+            calendarName: event.calendar.title,
+            description: event.description,
+            start: event.startDate,
+            stop: event.endDate,
+            color: UIColor(cgColor: event.calendar.cgColor).hex
```
heytherewill, line 85: Do we need these to be hexes now that are no longer constrained by the abstraction layer that was Xamarin?

sofokli1:
@juanlaube can you also take a look at the last commit that changes some cal entities from hex strings to uicolor?

PR #566, opened 2020-07-09
Fix TE view model selector

### LINKED ISSUE #565: Optimize timeEntryViewModelsSelector
## Description
The `timeEntryViewModelsSelector` is reacting on every state change by taking all TEs, mapping them into TEViewModels, and then and only then, diffing that array and discarding it if there are no changes.

This means that for every state change in the Timer that whole process runs. When there are a lot of time entries it results in a huge performance hit.

The way to fix this is to select on the TEs first, with `distinctUntilChanged` and then do the mapping.

## Definition of done
- [ ] The selector is fixed, performance is good.

Odrakir:
## Info
As described in the issue this fixed the huge performance issues in that code.

### Technical considerations
There's probably a better way to code selector that takes these cases in consideration, because this can easily happen again or be already happening somewhere.

### Relationships
Closes #565 

### Pull Request checklist

### Pointers
The code is very simple, but if you want to test it, try the previous version with a lot of TEs (like the ones generated [here](https://github.com/toggl/ios/pull/549)) and then try with these changes in place.

### Review template

- [ ] The code considers i18n

## Merge permissions
Go!


Odrakir:
That's in the `select` method in the store. It'll diff whatever you send it, that's why we have to first select the entities and then map. If we map inside the select (which is what we were doing) it'll incur in the performance problems.

PR #573, opened 2020-07-10
Use selected calendars

### LINKED ISSUE #455: Use selected calendars
## Description

We need to store the selected calendars in user defaults and use that when fetching events.

## Dependencies
Depends on #453 

## Definition of done
- [x] Selected calendars are stored in user defaults
- [x] All calls to `getCalendarEvents` use the selected calendars only

sofokli1:
### Summary
- does what the issue requires
- ports the userDefaults wrapper from the watch app and makes to force the predefined keys

Looking for some feedback regarding the last one

### Relationships
Closes #455

### Pull Request checklist

### Pointers
commit by commit

### Review template
- [ ] The code considers i18n

## Merge permissions
🦑


Modules/Utils/Utils/UserDefault/UserDefaultWrapper.swift
```diff
@@ -0,0 +1,49 @@
...
+    public var wrappedValue: T {
+        get {
+            guard let value = userDefaults.object(forKey: key.rawValue) else {
+                return defaultValue
+            }
+
+            return value as? T ?? defaultValue
+        }
+        set {
+            if let value = newValue as? OptionalProtocol, value.isNil() {
```
juanlaube, line 24: Do you really need the `OptionalProtocol`? Why not something like this?
```suggestion
            if let value = newValue as? Optional<T>, value == nil {
```

Modules/Utils/Utils/UserDefault/UserDefaultWrapper.swift
```diff
@@ -0,0 +1,49 @@
...
+                userDefaults.removeObject(forKey: key.rawValue)
+            } else {
+                userDefaults.set(newValue, forKey: key.rawValue)
+            }
+        }
+    }
+}
+
+public protocol OptionalProtocol {
```
juanlaube, line 33: 👀 I don't like this protocol, it's confusing and things that are not optional can conform to it, smells like bugs.

Modules/Utils/Utils/UserDefault/UserDefaultWrapper.swift
```diff
@@ -0,0 +1,49 @@
+import Foundation
+
+@propertyWrapper
```
juanlaube, line 3: 😍

Modules/Utils/Utils/UserDefault/UserDefaultWrapper.swift
sofokli1: this throughs a warning

Modules/Utils/Utils/UserDefault/UserDefaultWrapper.swift
sofokli1: I made it private so that it can't be used elsewhere

Modules/CalendarService/CalendarService/CalendarService.swift
```diff
@@ -43,12 +45,11 @@ public class DefaultCalendarService: CalendarService {
         return Single.just(calendars)
     }
 
-    public func getCalendarEvents(between start: Date, and endDate: Date, in calendars: [CalendarSource]) -> Single<[CalendarEvent]> {
+    public func getCalendarEvents(between start: Date, and endDate: Date) -> Single<[CalendarEvent]> {
```
Odrakir, line 48: Maybe this doesn't need to return a Single at all if we are not doing anything async.

Modules/CalendarService/CalendarService/CalendarService.swift
```diff
@@ -7,12 +7,14 @@ public protocol CalendarService {
...
     func requestAuthorization() -> Single<Bool>
     func getAvailableCalendars() -> Single<[CalendarSource]>
-    func getCalendarEvents(between start: Date, and endDate: Date, in calendars: [CalendarSource]) -> Single<[CalendarEvent]>
+    func getCalendarEvents(between start: Date, and endDate: Date) -> Single<[CalendarEvent]>
 }
 
 public class DefaultCalendarService: CalendarService {
 
     private let eventStore = EKEventStore()
+    @UserDefault(UserDefaultKey.selecteduserCalendarIds, defaultValue: [])
```
Odrakir, line 16: I still think user defaults are better as its own service. What's the reasoning to do it like this?

Modules/Models/Models/UserPreferences.swift
```diff
@@ -10,22 +10,19 @@ public struct UserPreferences: Equatable {
     public private(set) var selectedWorkspaceId: Int64
     public private(set) var smartAlertsState: SmartAlertsState
     public private(set) var useManualMode: Bool
-    public private(set) var useCalendarIntegration: Bool
```
Odrakir, line 13: Are we not including this in settings? This won't persist between installs. What's the behavior in the current app?

Modules/Settings/Settings/Common/SettingsState.swift
```diff
@@ -6,9 +6,11 @@ import Architecture
 public struct SettingsState: Equatable {
     public var route: RoutePath
     public var preferences: Loadable<UserPreferences>
+    public var userDefaults: AuroraUserDefaults
```
Odrakir, line 9: In my opinion user defaults is not state, but a dependency.

Modules/CalendarService/CalendarService/CalendarService.swift
sofokli1: I misunderstood your initial review. Should it also have a project of its own? (like the calService)

Modules/Models/Models/UserPreferences.swift
sofokli1: it is in `IUserPreferences` in the old app

Modules/CalendarService/CalendarService/CalendarService.swift
Odrakir: I think this can probably go to `OtherServices`

Modules/Models/Models/UserPreferences.swift
Odrakir: I mean, we are saving it to the DB, right?

sofokli1: it looks like we are saving it to keyValue storage. will move it to the userdefaults

Modules/TogglTrack/TogglTrack/AnalyticsReducer.swift
```diff
@@ -79,7 +82,7 @@ extension TimeEntriesLogAction {
 }
 
 extension LogSuggestionAction {
-    public func toEvents(_ state: AppState) -> [Event] {
+    public func toEvents(_ state: AppState, userDefaultService: UserDefaultService) -> [Event] {
```
Odrakir, line 85: I think this is not called from anywhere

Modules/TogglTrack/TogglTrack/AnalyticsReducer.swift
```diff
@@ -100,7 +103,7 @@ extension LogSuggestionAction {
             ]
 
             var calendarProviderState = CalendarSuggestionProviderState.unauthorized
-            if state.userDefaults.calendarPermissionWasGranted {
+            if userDefaultService.calendarPermissionWasGranted {
```
Odrakir, line 106: Using a service here feels a little odd to me, wouldn't it be better to add something to the calendar state that lets us know if it was authorised or not?

sofokli1: personally I am fine with this for now. if we find issues with this we can alway change it later on.

PR #591, opened 2020-07-15
Add date picker actions

### LINKED ISSUE #464: Add the day selected action
## Description
Add the day selected action to the date picker module. This action receives a date as a parameter.

## Dependencies
Depends on #463

## Definition of done
- [x] The selected date is updated in the reducer

### LINKED ISSUE #465: Add the calendar day swiped action
## Description
Add the calendar day swiped action to the date picker module. This action receives a swipe direction, either left or right.

## Dependencies
Depends on #463 

## Definition of done
- [x] When swiping right, the selected date is updated to the next day in the reducer
- [x] When swiping left, the selected date is updated to the previous day in the reducer
- [x] Doesn't update the selected date if the next or previous dates are not in the available dates*

*The available dates are: Two weeks in the past, starting with the current date.

### LINKED ISSUE #466: Add the week swiped action
## Description
Add the calendar week swiped action to the date picker module. This action receives a swipe direction, either left or right.

## Dependencies
Depends on #463

## Definition of done
- [x] When swiping right, the selected date is updated to the same day in the next week in the reducer
- [x] When swiping left, the selected date is updated to the same day in the previous week in the reducer
- [ ] ~Doesn't update the selected date if the new values are not in the available dates~
- [x] When swiping right and the new selected date would end up not in the available dates, selects the last available date
- [x] When swiping left and the new selected date would end up not in the available dates, selects the first available date

sofokli1:
### Summary
- Covert date picker from a module to a feature
- Add the 3 actions from the issues
- Add tests for those
- Update the reducer test template to work out of the box in the future

### Relationships
Closes #464
Closes #465
Closes #466

### Pull Request checklist

### Pointers
Check the state, action reducer and tests mainly.

### Review template
- [ ] The code considers i18n

## Merge permissions
🦑


Modules/Calendar/Calendar/DatePicker/DatePickerAction.swift
```diff
@@ -0,0 +1,23 @@
...
+extension DatePickerAction: CustomDebugStringConvertible {
+
+    public var debugDescription: String {
+        switch self {
+        case let .daySelected(date):
+            return "DaySelected: \(date)"
+        case let .weekStripeSwiped(direction):
+            return "WeekStripeSwiped: \(direction == .left ? "left" : "right")"
+        case let .calendarSwiped(direction):
+            return "CalendarSwiped: \(direction == .left ? "left" : "right")"
```
juanlaube, line 20: You don't need this string interpolation, Swift is smart 😄 

```suggestion
        case let .weekStripeSwiped(direction):
            return "WeekStripeSwiped: \(direction)"
        case let .calendarSwiped(direction):
            return "CalendarSwiped: \(direction)"
```

PR #592, opened 2020-07-16
Refactor/reorganize state

Odrakir:
## Info
This PR attempts to reorganise state so it's easier to reason about it. It also adds some code to help manage it.

### Summary
This is the new structure of the state. The properties between parenthesis indicate that those are not owned but that part of the state, but they come from other parts (not necessarily the parent). For example, the entities don't have to live on the root AppState, the way I see it they belong to the Timer, so that's where I put them, they can be accessed from other parts of the state still.

```
# AppState
user: Loadable<User>
route: RoutePath

  ## LoadingState
  (entities, route)
  
  ## OnboardState
  (user, route)
  email: String
  password: String
  
  ## TimerState
  (user, calendarEvents)
  entities: TimeLogEntities      

    ### TimeEntriesLogState
    (logSuggestions, editableTimeEntry, entities)
    expandedGroups: Set<Int>
    entriesPendingDeletion: [TimeEntry.Identifier]

    ### LogSuggestionsState
    (entities, calendarEvents, user)
    logSuggestions: [LogSuggestion]    

    ### StartEditState
    (user, entities)
    autocompleteSuggestions: [AutocompleteSuggestion]
    dateTimePickMode: DateTimePickMode
    cursorPosition: Int
    editableTimeEntry: EditableTimeEntry?
    
    ### RunningTimeEntryState
    (user, entities, editableTimeEntry)

    ### ProjectState
    (editableTimeEntry.editableProject, entities.projects)

  ## CalendarState
  (user, entities)
  editableTimeEntry: EditableTimeEntry? // This one is different from the one in the Timer
  calendarEvents: [String: CalendarEvent]
  selectedItem: Either<EditableTimeEntry, CalendarEvent>?
  
    ### CalendarDayState
    (user, entities, calendarEvents, selectedItem)
    selectedDate: Date

    ### ContextualMenuState
    (user, selectedItem, editableTimeEntry, entities)

  ## SettingsState
  (route)
  preferences: Loadable<UserPreferences>
```

To tidy up everything, when a part of the state needs some local state it'll go into a `Local` struct defined inside the substate. Also, I added `@dynamicMemberLookup` where necessary so it's possible to access that local state directly as if it was not embedded inside the `Local` struct (`myState.myProperty` instead of `myState.local.myProperty`)

There might be better ways to do this. I'm still not 100% sure, but I think this is better than what we have right now.

### Pull Request checklist

### Pointers
Some of the states are simpler than others (settings, onboarding), you can start with those commits to have a better understanding of what's going on and then move to the bigger ones.

### Review template

- [ ] The code considers i18n

## Merge permissions
🚫  Let me do it


Modules/Calendar/Calendar/CalendarDay/CalendarDayState.swift
```diff
@@ -4,10 +4,32 @@ import Models
 import Utils
 import Timer
 
-public struct CalendarDayState: Equatable {
+@dynamicMemberLookup
```
juanlaube, line 7: 😍

Odrakir:
I thought about using the protocol for `@dynamicMemberLookup`, but I'm afraid it would make things a little harder to understand, and since this is only used in a few places in the app I'm fine with the repetition.

We can reconsider it later down the line.

Modules/Calendar/Calendar/Common/CalendarState.swift
```diff
@@ -4,64 +4,68 @@ import Models
...
     }
-
+    
     internal var contextualMenuState: ContextualMenuState {
         get {
             ContextualMenuState(
                 user: user,
-                selectedItem: localCalendarState.selectedItem,
-                editableTimeEntry: editableTimeEntry,
+                selectedItem: local.localCalendarDayState.selectedItem,
```
dshr, line 60: This feels a bit awkward to me, maybe there's a way to avoid accessing things through `localCalendarDayState` here?

Odrakir: That one should actually be `calendarDayState.selectedItem`, which under the hood is the same, but a little cleaner.

I have to say I'm not to familiar with the Calendar logic, I don't really know if selectedItem really belongs in CalendarDayState, as in logically it's the part of the state that owns it. Maybe it's better at the root of the CalendarState?

juanlaube: I think the `selectedItem` should be in the local state at the root level, it's shared between the calendar day and the contextual menu, having it there is much cleaner in my opinion.

Odrakir:
I've updated the calendar state moving `selectedItem` to the root level.

I've also changed `entriesPendingDeletion` to be an array instead of a set, it makes sense to keep the order of deletion and it'll help with tests in my next PR.

PR #593, opened 2020-07-17
 Add SendFeedbackTapped action 

### LINKED ISSUE #432: Add SendFeedbackTapped action
## Description
Add SendFeedbackTapped action. It should update the `UserPrefs` object in the state. Delete the message from the state. Add the missing property if needed.

Also add the `SendFeedback` effect and `FeedbackSent` action. Mock what can't be implemented like the actual network call and make an issues for that to be fixed later on

## Dependencies
#417
## Definition of done
- [ ] Add SendFeedbackTapped action
- [ ] Add SendFeedback effect
- [ ] Add FeedbackSent action
- [ ] Mock and make issues for what is not possible to implement
- [ ] Add tests

dshr:
### Summary
This PR adds all the feeback-sending actions, the relevant API call and dependency handling.

### Relationships
Closes #432 

### Pull Request checklist

### Pointers
Commit by commit FTW

### Review template

## Merge permissions
Once done, and the new issues are created (they will reference this PR), only then.


Modules/Settings/Settings/Common/SettingsAction.swift
```diff
@@ -17,6 +17,10 @@ public enum SettingsAction: Equatable {
     case durationFormatSelected(DurationFormat)
     case dateFormatSelected(DateFormat)
     case use24HourClockToggled
+    case sendFeedbackTapped(String)
+    case feedbackSent
+    case setSendFeedbackError(ErrorType)
+    case sendFeedbackResultSeen
```
Odrakir, line 23: What's this action for? Is not in Notion and it's not being sent right now.

Modules/Settings/Settings/Common/SettingsReducer.swift
```diff
@@ -147,3 +182,42 @@ func requestNotificationsPermission(_ notificationsPermissionsService: Notificat
...
+// swiftlint:disable todo
+func assembleFeedbackData(repository: SettingsRepository,
+                          preferences: Loadable<UserPreferences>,
+                          time: Time) -> Single<FeedbackData> {
+    guard let usingManualMode = preferences.value?.useManualMode
+        else { fatalError("No user preferences present") }
+    return Observable
+        .combineLatest(
+            repository.getTimeEntriesCount().asObservable(),
+            repository.getWorkspacesCount().asObservable(),
```
Odrakir, line 195: Why are we not getting these from the state?

Modules/Settings/Settings/Common/SettingsReducer.swift
```diff
@@ -147,3 +182,42 @@ func requestNotificationsPermission(_ notificationsPermissionsService: Notificat
...
+            resultSelector: { timeEntriesCount, workspacesCount in
+                return FeedbackData(
+                    accountTimeZone: nil, // TODO: get actual time zone
+                    numberOfWorkspaces: workspacesCount,
+                    numberOfTimeEntries: timeEntriesCount,
+                    numberOfUnsyncedTimeEntries: 0, // TODO: get actual number of unsynced entries
+                    numberOfUnsyncableTimeEntries: 0, // TODO: get actual number of unsyncable entries
```
Odrakir, line 202: I think we should be creating issues, not adding "TODO"

Modules/Settings/SettingsTests/mocks/MockFeedbackAPI.swift
```diff
@@ -0,0 +1,57 @@
...
+import RxSwift
+
+struct MockFeedbackError: Error {
+    
+}
+
+class MockFeedbackAPI: FeedbackAPI {
+    
+    let mockRepository: MockSettingsRepository
+    let mockTime: Time
```
Odrakir, line 14: Are these dependencies really necessary? If this is a mock it should be as simple as possible, returning dummy values that can be set before hand and not calling other mocks if it can be avoided.

Modules/TogglTrack/TogglTrack/App/AppEnvironment.swift
```diff
@@ -34,21 +39,38 @@ public struct AppEnvironment {
...
+            else { fatalError("No version and build number") }
+        return AppEnvironment(
+            api: API.local,
+            repository: Repository.local,
+            time: Time(),
+            schedulerProvider: SchedulerProvider(),
+            analytics: CompositeAnalyticsService(analyticsServices: []),
+            calendarService: DefaultCalendarService(),
+            notificationsService: NotificationsService(),
+            platformInfo: PlatformInfo(
```
Odrakir, line 58: Can we move this up to a local variable so the AppEnvironment initialiser is cleaner?

Modules/TogglTrack/TogglTrack/App/AppEnvironment.swift
```diff
@@ -57,6 +79,14 @@ public struct AppEnvironment {
             analytics: CompositeAnalyticsService(analyticsServices: analyticsServices),
             calendarService: DefaultCalendarService(),
             notificationsService: NotificationsService(),
+            platformInfo: PlatformInfo(
```
Odrakir, line 82: Can we move this up to a local variable so the AppEnvironment initialiser is cleaner?

Modules/TogglTrack/TogglTrack/App/AppReducer.swift
```diff
@@ -38,6 +39,14 @@ func createAppReducer(environment: AppEnvironment) -> Reducer<AppState, AppActio
...
-            .pullback(state: \.loadingState, action: \.load)
+            .pullback(state: \.loadingState, action: \.load),
+        createSettingsReducer(repository: environment.repository,
+                              notificationsPermissionsService: environment.notificationsService,
+                              time: environment.time,
+                              calendarService: environment.calendarService,
+                              platformInfo: environment.platformInfo,
+                              feedbackAPI: environment.feedbackAPI,
+                              userDefaultsService: environment.userDefaultService)
+            .pullback(state: \.settingsState, action: \.settings)
```
Odrakir, line 50: While on it can you fix the indentation in this whole reducer combining? Make it a bit more consistent

Modules/Settings/Settings/Common/SettingsReducer.swift
```diff
@@ -3,14 +3,18 @@ import Architecture
 import Models
 import RxSwift
 import Repository
+import API
 import OtherServices
 import CalendarService
 
-// swiftlint:disable cyclomatic_complexity function_body_length
-func createSettingsReducer(
+// swiftlint:disable cyclomatic_complexity function_body_length function_parameter_count
```
Odrakir, line 10: Not for this PR, but I think it's time to consider moving the environment back to the store. First because it's getting huge, too many params and second because if we'll end up using TCA, that's how they do it.

Modules/Settings/Settings/Common/SettingsAction.swift
dshr: I looked at the droid implementation for ideas, and it's used to hide the "feedback sent" message.

Modules/Settings/Settings/Common/SettingsReducer.swift
dshr: <image src="https://media.giphy.com/media/3o85xIO33l7RlmLR4I/giphy.gif" />

Modules/Settings/SettingsTests/mocks/MockFeedbackAPI.swift
dshr: It makes the tests a bit more robust IMO, by not having us remember all the pre-set values. I'm okay doing things either way though.

Modules/Settings/Settings/Common/SettingsReducer.swift
dshr: Kind of felt more proper to me. Will the unsynced and unsyncable TEs also be in the state as well?

Modules/Settings/SettingsTests/mocks/MockFeedbackAPI.swift
dshr: Removed these to make things a bit cleaner, good point

PR #594, opened 2020-07-17
Refactor architecture

Odrakir:
## Info
This PR refactors certain aspects of the architecture with the following purpose:
- Make adding subscriptions easier
- Adapt our code to the original architecture so we are ready when we decide to migrate (either if we use SwiftUI/Combine or not)
- Finally by following the original architecture more close we'll avoid certain mistakes

### Summary
This is a huge PR because the architecture changes affect parts in the whole application. But actually most of it is just renaming stuff or adopting the new API. The things that matter are in the first commit.
- First Effect is an Observable instead of containing a Single. This allows for long running effects fro subscriptions. It also adds some methods to merge and concatenate effects because...
- Reducers don't return arrays of effects anymore, they return just single effects. So if we want a reducer to result in multiple actions we need to merge or concatenate effects. (Also instead of `return []` we'll do `return .none` for now on, and to return an effect that immediately returns an action we'd do `Effect.just(ActionType.myAction)`)
- The code inside the Store changes to adapt to reducers not returning arrays, but the API stays more or less the same. The only thing I removed is you can't batch actions anymore, you need to send them one after another, but that's not a big deal.

The following commits after that one adapt the whole app to those changes.

And then finally in `✅ reducer assert and debug` I adapt the reducer assert code to these changes as well, and while on it I add some code to make debugging test errors easier (now it prints diffing). Also the syntax changes a little:
- To create a step you do `.send(ActionType.myAction)` instead of `Step(.send, ActionType.myAction)`
- The reducer assert doesn't take a scheduler anymore. Instead you add a `do { self.scheduler.start() }` to advance the test scheduler. The `do` step could do anything else besides that.

### Relationships
As mentioned above this will help with implementing subscriptions

### Pull Request checklist

### Pointers
Take a look at the summary for a guide on how to review

### Review template

- [ ] The code considers i18n

## Merge permissions
🚫 Let me do it


dshr:
I kind of dread having to merge my branch after this gets merged though..

Odrakir:
I can help fixing conflicts, I already did it for the rest of the app, so I know what needs to be done. Just let me know when, @dshr

PR #595, opened 2020-07-17
Subscriptions

Odrakir:
## Info
This PR adds a way to add long running effects, aka subscriptions. They basically are observables that are started/cancelled depending on the state of the app and emit actions.

This doesn't add subscriptions to handle repository changes, it just adds the architecture feature to add subscriptions for that or whatever later on.

### Summary
For subscriptions to work we first need to make effect cancellable, that's what the first commit does.

This is how you'd add a subscription, it's just another reducer that can be combined with other reducers at any level of the app: 

```swift
Reducer.subscriptions { state in
    struct TimerId: Hashable {}
    guard state.route.localizedCaseInsensitiveContains("reports") else { return [:] }
    return [
        TimerId(): Observable<Int>.interval(1, scheduler: MainScheduler.instance)
            .mapTo(AppAction.timerTick)
            .eraseToEffect()                    
    ]
}
```

### Pull Request checklist

### Pointers
Not much to say, you can test this with the code above.

### Review template

- [ ] The code considers i18n

## Merge permissions
Go!


semanticer:
We implemented this slightly differently, I wanted to keep our effect single emit only so we can stick with suspended functions, this makes effect code simple "synchronous" and I also like the fact that   I will leave the real review for people who are able to comment on the details of your implementation.

semanticer:
We implemented this slightly differently, I wanted to keep our effects single emit only so we can stick with suspended functions, this makes effect code simple and synchronous-like. That's why I introduced subscriptions as a separate concept, they are basically another thing that you can plugin into your store. However, I think your approach makes complete sense when you are already using observables for effects. I'm not sure how consistent we need to be here, I'm just stating the facts. 
I'm gonna leave the real review for people who are actually able to comment on the details of your implementation.

Draft Android PR is here btw https://github.com/toggl/android/pull/602

Modules/Architecture/Architecture/Core/Effect.swift
```diff
@@ -134,3 +134,93 @@ extension ObservableConvertibleType {
...
+        _dispose()
+    }
+
+    static func == (lhs: AnyDisposable, rhs: AnyDisposable) -> Bool {
+        return ObjectIdentifier(lhs) == ObjectIdentifier(rhs)
+    }
+
+    func hash(into hasher: inout Hasher) {
+        hasher.combine(ObjectIdentifier(self))
+    }
```
dshr, line 156: Could you re-enable `identifier_name` after this class?

Modules/Architecture/Architecture/Core/Effect.swift
```diff
@@ -134,3 +134,93 @@ extension ObservableConvertibleType {
...
+}
+
+extension Effect {
+    public func cancellable(id: AnyHashable, cancelInFlight: Bool = false) -> Effect {
```
dshr, line 160: What's the use case for `cancelInFlight`?

Odrakir: According to the original architecture: "Determines if any in-flight effect with the same identifier should be canceled before starting this new one."

PR #597, opened 2020-07-22
Login API

### LINKED ISSUE #568: Email/password API interaction
## Description
Implement the login API call. You can look at the Xamarin app to see how the request should be serialized. In the iOS app there seems to be some implementation already. Validate it is correct and add tests.
## Definition of done
- Login API can be called with email/password
- The response is handled
- Tests are added

Odrakir:
## Info
Cleans up the login API and adds tests.

### Summary
This was more or less done, I just refactored it a little so it's a little bit more solid and added tests.

### Relationships
Closes #568 

### Pull Request checklist

### Pointers
Not much to say, easy stuff

### Review template

- [ ] The code considers i18n

## Merge permissions
Do it!


dshr:
But fix @heytherewill's request first

PR #598, opened 2020-07-22
Add logout logic

### LINKED ISSUE #436: Add SignOutTapped action
## Description
Add SignOutTapped action, `SignOut` effect and `SignOutCompleted` action. 
SignOut effect and the SignOutCompleted action are the most interesting part here since it should change the state in a way that it takes the user to the login view and triggers a clean up of the data like TEs etc. and api tokes and what not. Whatever is not possible to implement at this point, leave a comment and make an issue so that it can be handled later on.

As for the correct name `SignOut` vs `Log Out` I leave it to the dev working on this to talk with product to find out what is best.

## Dependencies
#417
## Definition of done
- [ ] Add SignOutTapped action
- [ ] Add SignOut effect
- [ ] Add SignOutCompleted action
- [ ] Add tests

Odrakir:
## Info
Adds actions, reducer logic and service methods to successfully logout user.
### Technical considerations
I decided to move this to the `AppReducer`, as it makes sense to take it as a global level action even if it's started in Settings (although it could also be started as a response to an API error, so that's more reason to move it there). It might need access to a lot of services and it needs to reset most (if not all) the state

I'm willing to hear other thoughts though.

Also, this is not clearing everything we'll need to clear in the end, we'll need to add more stuff to the reducer as needed.

### Relationships
Closes #436 

### Pull Request checklist

### Pointers
Tiny commits are easy to review commits

### Review template

- [ ] The code considers i18n

## Merge permissions
Do it


Modules/OtherServices/OtherServices/UserDefaultService.swift
```diff
@@ -60,4 +64,9 @@ public class UserDefaultService {
 
     @UserDefault(UserDefaultKey.calendarPermissionWasGranted, defaultValue: false)
     public var calendarPermissionWasGranted: Bool
+
+    public func clear() {
+        _useCalendarIntegration.clear()
+        _selectedUserCalendarIds.clear()
```
sofokli1, line 70: why not the `calendarPermissionWasGranted`?

Odrakir: I'm not sure about that one, to be honest. We can't clear the system status on that permission, so I guess we have to still keep track of wether the user gave permission already or not.

sofokli1: I guess we can change it later if needed

Modules/Database/Database/CoreData/CoreDataModel.swift
```diff
@@ -62,8 +62,16 @@ extension CoreDataModel {
     static func deleteAll(
         in context: NSManagedObjectContext
     ) throws {
-        let fetchRequest: NSFetchRequest<NSFetchRequestResult> = NSFetchRequest(entityName: ManagedObject.entityName)
-        let deleteRequest = NSBatchDeleteRequest(fetchRequest: fetchRequest)
-        try context.persistentStoreCoordinator?.execute(deleteRequest, with: context)
+        #if DEBUG //Batch delete doesn't work with the in-memory store used in tests
```
sofokli1, line 65: isn't this going to be slow when testing logout in simulator/device during development?

Odrakir: I don't think it will, because it will happen in memory, but we'll keep an eye on that.

PR #653, opened 2020-07-27
Add ForgotPasswordTapped action and wire up the button

### LINKED ISSUE #569: Add ForgotPasswordTapped action and wire up the button
## Description
Add ForgotPasswordTapped action and wire up the button

## Definition of done

- Action is added and reduced
- Tests are added
- Button is added to the login view
- Action is emitted  

dshr:
## Info
This PR adds the ForgotPasswordTapped action, refactors the login view a bit (we're using a stackview there to handle the showing/hiding of the forgot password button that is to be added down the line), adds the forgot password button and connects it to the new action.

### Relationships
Closes #569 

### Pull Request checklist

### Pointers
Commit by commit FTW

### Review template
- [ ] The code considers i18n

## Merge permissions
Go for it!


Modules/Utils/Utils/UI/TextFormatting/PartiallyUnderlinedString.swift
```diff
@@ -0,0 +1,15 @@
+import UIKit
+import Foundation
+
+public func partiallyUnderlinedString(segments: [(_: String, underline: Bool)]) -> NSAttributedString {
+    let attributedString = NSMutableAttributedString()
+    segments.forEach { segment in
+        let attributes: [NSAttributedString.Key: Any]? = segment.underline
+            ? [.underlineStyle: NSUnderlineStyle.single.rawValue]
+            : nil
+        let part = NSAttributedString.init(string: segment.0,
```
sofokli1, line 10: why not `segment.text` or `.string`

Modules/Onboarding/Onboarding/EmailLogin/Views/LoginViewController.swift
```diff
@@ -81,6 +91,16 @@ public class LoginViewController: UIViewController, Storyboarded {
         store.dispatch(.passwordEntered("alskjdaslkjaskdj"))
         store.dispatch(.loginTapped)
     }
+    
+    private func setUpTitles() {
+        let forgotPasswordQuestion = "forgotPasswordQuestion".localized
```
sofokli1, line 96: shouldn't we add this strings to `Assets > Extensions > Strings.swift` so that the raw format is only use there and make it harder to make mistakes?

Modules/Utils/Utils/UI/TextFormatting/PartiallyUnderlinedString.swift
dshr: My idea was to have it like:
``` Swift
let passwordResetButtonTitle = partiallyUnderlinedString(segments: [
    (forgotPasswordQuestion + " ", underline: false),
    (forgotPasswordCTA, underline: true)
])
```
instead of 
``` Swift
let passwordResetButtonTitle = partiallyUnderlinedString(segments: [
    (text: forgotPasswordQuestion + " ", underline: false),
    (text: forgotPasswordCTA, underline: true)
])
```

As it turns out, I can drop tuple label if I want to, so it works alright.

Modules/Onboarding/Onboarding/EmailLogin/Views/LoginViewController.swift
dshr: Completely forgot about that one. Gotta note it somewhere so we can add a linter for it. Or maybe user SwiftGen for it? 🤔

Modules/Assets/Assets/fr.lproj/Localizable.strings
```diff
@@ -16,3 +16,5 @@
 "createTag" = "Créer label";
 "createProject" = "Créer projet";
 "previouslyTracked" = "Previously tracked Time Entries";
+"forgotPasswordQuestion" = "Forgot Password?";
```
simonrozsival, line 19: Don't we already have strings for these phrases in the Xamarin app already?

dshr: Not exactly the same phrases, similar ones only. Like, I'm not confident enough to extract the proper infinitives from them for "Reset".

PR #656, opened 2020-07-28
Add new fake data

Odrakir:
## Info
This changes the text files with fake data. It adds:
130 Clients
630 Projects
308 Tasks
500 Tags
17970 Time entries

This will allow us to test performance with big numbers.

### Technical considerations
The new data will be inserted in the DB on new installs, if you want to have less entries it can be done by modifying the API or Repository to take only the first x entries.

The data was generated with the Workspace Generator, but since API endpoints only allow access to the last 3 months of data everything is squashed in the last 3 months, which results in a bunch of entries per day, this is not very realistic, but it'll serve testing purposes for now.

### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
🚫 Don't!


Odrakir:
I'm closing this one since this completely changed and we should find another way of populating the DB with dummy data (if needed)

PR #657, opened 2020-07-29
Add the forgot password UI

### LINKED ISSUE #570: Add forgot password UI
## Description
Add the UI needed for forgot password. If the relevant actions for the view are already implemented, wire them up.
## Definition of done
- View is added
- Actions are wired up, if they are available

### LINKED ISSUE #575: Forgot password - add CloseButtonTapped action
## Description
Add the action handling closing the forgot password view. Wire it up to the UI if it’s there.
## Definition of done
- Action is added and reduced
- It’s wired with the UI if the UI is there
- Tests are added

dshr:
## Info
This PR adds the forgot password UI, adds it's `closeButtonTappedAction` and connects it to the view

### Relationships
Closes #570
Closes #575 

### Pull Request checklist

### Pointers
Commit by commit should work best

### Review template
- [ ] The code considers i18n

## Merge permissions
Go for it!


Modules/Onboarding/Onboarding/PasswordReset/PasswordResetCoordinator.swift
```diff
@@ -0,0 +1,22 @@
...
+    private var store: Store<OnboardingState, PasswordResetAction>
+        
+    public init(store: Store<OnboardingState, PasswordResetAction>) {
+        self.store = store
+    }
+    
+    public override func start() {
+        let viewController = PasswordResetViewController.instantiate()
+        viewController.store = store
+        navigationController.isNavigationBarHidden = true
```
Odrakir, line 15: Why are you using a `NavigationCoordinator` at all if you hide the bar and don't ever push a new VC?

Also... this is design, but, shouldn't we be pushing this view from the `EmailLoginViewController` instead of showing it as a modal? Can you check with design? In my opinion there's too many modals already.

dshr: Good point re the navigation coordinator, I derped and just did the same thing as the `EmailLoginCoordinator`. As for the modal-ness of the password reset view — I think the big X is a giveaway that it's not pushed onto a navigation controller's stack.

Modules/Onboarding/Onboarding/PasswordReset/Views/PasswordResetViewController.swift
```diff
@@ -0,0 +1,59 @@
...
+            (passwordResetHelpQuestion + " ", false),
+            (passwordResetHelpCTA, true)
+        ])
+        contactUsButton.setAttributedTitle(contactUsButtonTitle, for: .normal)
+    }
+}
+
+extension PasswordResetViewController: UIAdaptivePresentationControllerDelegate {
+    public func presentationControllerDidDismiss(_ presentationController: UIPresentationController) {
+        store.dispatch(.closeButtonTapped)
```
simonrozsival, line 57: This code never runs. I put a breakpoint here and it wasn't hit.

simonrozsival: I suggest simply calling `self.dismiss` in the `rx.tap` handler and then overriding `viewDidDisappear` and dispatching the close action there.

dshr: The problem wasn't there, but with settings `UIAdaptivePresentationControllerDelegate`. Good catch still, I'm gonna push a fix in a sec

Modules/Onboarding/Onboarding/PasswordReset/PasswordResetCoordinator.swift
Odrakir: No, I know the design puts this screen as a modal, I'm wondering why and if it wouldn't make more sense as a pushed screen, given that we are already in a modal.

PR #663, opened 2020-08-03
Performance optimizations

Odrakir:
## Info
This PRs adds some performance optimizations to the way the store handles state and some other things in the log.

### Summary
Basically this:
- Refactors  the `select` methods in store to use key paths instead of functions, this prevents us from doing a lot of mapping before the `distinctUntilChanged`.
- Converts most of `TimeEntryViewModel` properties into computed properties so they are only calculated when needed.
- Makes some optimizations to the comparison of items in the log.

### Technical considerations
I've been trying the app with the Instruments and a release build and performance seems fine even with a lot of TEs (17.000). Still there might be some optimizations we might want to do in the future, like paging, which was drafted and discussed and it'd be pretty easy to do.

### Relationships
Doesn't close anything, but it's related to the performance OKR

### Pull Request checklist

### Pointers
One commit at a time

### Review template

- [ ] The code considers i18n

## Merge permissions
Go!


dshr:
The tests still need some ♥️ though

Modules/Timer/Timer/TimeEntriesLog/Views/TimeEntriesLogViewController.swift
```diff
@@ -83,6 +83,7 @@ public class TimeEntriesLogViewController: UIViewController, Storyboarded {
                 store.select(\.entriesPendingDeletion),
                 resultSelector: toSectionsMapper
             )
+                .debounce(.milliseconds(100))
```
heytherewill, line 86: I feel like `sample` works best for here, but I could be wrong

Odrakir: I think it could be problematic if the sampling happens right after one of the observables emitting and right before another one. It's an edge case and very unlikely, but debounce is safer.

PR #664, opened 2020-08-03
Make the Password Reset view scroll with the keyboard

### LINKED ISSUE #659: Make the Password Reset view scroll with the keyboard
## Description
Scroll the whole view up whenever the keyboard is open, and scroll it back when it's hidden.

## Dependencies
#657 

## Definition of done
- [ ] Opening the keyboard scrolls the view, making the contact us button visible
- [ ] Closing the keyboard scrolls the view back to it's initial position

dshr:
## Info
This PR makes the password reset UI scroll with the keyboard, along with introducing a useful protocol for easier keyboard interaction down the line.

### Relationships
Closes #659 

### Pull Request checklist

### Pointers
Commit by commit!

### Review template
- [ ] The code considers i18n

## Merge permissions
Go for it!


dshr:
`IQKeyboardManager` is cool and all, but IMO it will make our lives quite difficult in our bottom sheet-based views, while this approach allows us to define our own scrolling strategy on a per-screen basis.

Odrakir:
Yes please, we don't need a third party dependency to handle the keyboard.

Modules/Onboarding/Onboarding/PasswordReset/Views/PasswordResetViewController.swift
```diff
@@ -52,6 +58,34 @@ public class PasswordResetViewController: UIViewController, Storyboarded {
...
+    }
+    
+    public func keyboardWillChangeFrame(_ notification: Notification) {
+        guard let keyboardSize = notification.keyboardSize else { return }
+        setScrollViewInsets(keyboardSize.height)
+    }
+    
+    public func keyboardWillHide(_ notification: Notification) {
+        scrollView.contentInset = .zero
+        scrollView.scrollIndicatorInsets = .zero
```
sofokli1, line 79: why not just `setScrollViewInsets(0)`

Modules/Onboarding/Onboarding/PasswordReset/Views/PasswordResetViewController.swift
```diff
@@ -52,6 +58,34 @@ public class PasswordResetViewController: UIViewController, Storyboarded {
     }
 }
 
+extension PasswordResetViewController: KeyboardObserving {
+    public func keyboardWillShow(_ notification: Notification) {
+        guard let keyboardSize = notification.keyboardSize else { return }
+        setScrollViewInsets(keyboardSize.height)
+        scrollView.scrollRectToVisible(CGRect(x: 0,
+                                              y: scrollView.contentSize.height,
```
sofokli1, line 66: overall the idea is good but this line makes this usable only here and hard to reuse. going to the bottom may be ok for this view but in other cases going to the bottom may kick the text input view outside of the visible rect and need manual scrolling from the user. No need to change anything here, just pointing it out

dshr: Yeah, this is only for this password reset view, thing will be different for other views.

Modules/Onboarding/Onboarding/PasswordReset/Views/PasswordResetViewController.swift
dshr: Good point

Modules/Onboarding/Onboarding/PasswordReset/Views/PasswordResetViewController.swift
```diff
@@ -52,6 +58,33 @@ public class PasswordResetViewController: UIViewController, Storyboarded {
...
+                                              width: 1,
+                                              height: 1),
+                                       animated: true)
+    }
+    
+    public func keyboardWillChangeFrame(_ notification: Notification) {
+        guard let keyboardSize = notification.keyboardSize else { return }
+        setScrollViewInsets(keyboardSize.height)
+    }
```
Odrakir, line 75: Are you sure we need this?

Modules/Utils/Utils/Extensions/Notification+KeyboardExtensions.swift
```diff
@@ -0,0 +1,12 @@
...
+import UIKit
+
+public extension Notification {
+    var keyboardSize: CGSize? {
+        return (userInfo?[UIResponder.keyboardFrameEndUserInfoKey] as? NSValue)?.cgRectValue.size
+    }
+    
+    var keyboardAnimationDuration: Double? {
+        return userInfo?[UIResponder.keyboardAnimationDurationUserInfoKey] as? Double
+    }
```
Odrakir, line 11: I think we are not using this one

Modules/Onboarding/Onboarding/PasswordReset/Views/PasswordResetViewController.swift
```diff
@@ -38,6 +40,10 @@ public class PasswordResetViewController: UIViewController, Storyboarded {
         self.presentationController?.delegate = self
     }
     
+    deinit {
+        unregisterFromKeyboardEvents()
+    }
```
Odrakir, line 45: Why `viewDidLoad` and `deinit` instead of `viewWillAppear` and `viewWillDisappear`? Not a big deal, I'm wondering if there's a reason for that. In this view it's fine, but some other view might need to stop receiving events when not visible.

Modules/Onboarding/Onboarding/PasswordReset/Views/PasswordResetViewController.swift
dshr: Yeah, there are cases for it — the easiest example is if you disable predictive text and switch between the regular keyboard and the emoji one. We also have people using custom keyboards as well, which also add to the complexity of things.

Modules/Utils/Utils/Extensions/Notification+KeyboardExtensions.swift
dshr: I wrote the extension as a bit of a future-proofing thing tbh, since there's definitely a case for animating the screen resize with the same duration as the keyboard appearing.

Modules/Onboarding/Onboarding/PasswordReset/Views/PasswordResetViewController.swift
dshr: No reason here, just generally this is the default mindset for me. I'll switch to `viewWillAppear` and `viewWillDisappear` for consistency's sake though.

PR #668, opened 2020-08-04
Reports totals api

### LINKED ISSUE #501: Create endpoint for report totals
The endpoint should take a user id, a workspace id, a start date and an end date. Endpoint is as follows:
`workspace/{workspaceId}/search/date_entries/totals`, with the payload being in the format of:
```
{
    "start_date": "2020-06-29",
    "end_date": "2020-07-05",
    "user_ids": [4674715],
    "with_graph": true
}
```

### LINKED ISSUE #503: Create a reports service
- [x] Create a separate service for reports fetching, inject the API into it
- [ ] Create a mock reports service for testing

### LINKED ISSUE #504: Create struct data type for report totals
## What needs to be done?
- [x] Create a `Resolution` enum with the following values:
```
    day,
    week,
    month
```
- [x] Create struct data type for report totals in the reports service. The format should be the following:
```
    startDate: Date
    endDate: Date?
    resolution: Resolution
    groups: [(total: TimeInterval, billable: TimeInterval)]
```
- [x] Look at the Xamarin implementation for some guidance
- [ ] Create a method that calls the API and parses the output of the mock endpoint into the new struct.

### Dependencies
Depends on #501 and #503

### LINKED ISSUE #510: Create the .reportsViewAppeared action
## What needs to be done?
- [x] Add the `.reportsViewAppeared` action, it should use the default start and end dates, and construct a `DateRangeSelection` with `SelectionSource` set to `.initial`
- [x] Emit the `LoadReportsEffect`, passing the `DateRangeSelection` to it
- [ ] Add appropriate tests
- [ ] Update the [docs](https://www.notion.so/toggl/Reports-54cac91332aa481db6b1de0d3b90aeb9)

### Dependencies
Depends on #511

### LINKED ISSUE #512: Create the .reportLoaded action
## What needs to be done?
- [x] Create the `.reportLoaded` action with `DateRangeSelection`, `ReportTotals` and `ReportSummary` parameters
- [ ] This action should emit a `ReportsSuccess` analytics event
- [x] This action set the `ReportTotals` and `ReportSummary` loadables to `.loaded`
- [x] Add appropriate tests
- [x] Update the [docs](https://www.notion.so/toggl/Reports-54cac91332aa481db6b1de0d3b90aeb9)

### Dependencies
Depends on #511

### LINKED ISSUE #513: Create the .reportFailed action
## What needs to be done?
- [x] Create the `.reportFailed` action with `DateRangeSelection` and `Error` parameters
- [x] This action should emit a `ReportsFailure` analytics event
- [x] This action set the `ReportTotals` and `ReportSummary` loadables to `.failed`
- [x] Add appropriate tests
- [ ] Update the [docs](https://www.notion.so/toggl/Reports-54cac91332aa481db6b1de0d3b90aeb9)

### Dependencies
Depends on #511

### LINKED ISSUE #514: Create the LoadReportsEffect
## What needs to be done?
- [x] This effect should take a `DateRangeSelection` as a parameter
- [x] This effect should call the reports service for `ReportTotals` and `ReportSummary`, and then emit the `.reportLoaded` or `.reportFailed` actions
- [ ] Add appropriate tests
- [ ] Update the [docs](https://www.notion.so/toggl/Reports-54cac91332aa481db6b1de0d3b90aeb9)

### Dependencies
Depends on #511

heytherewill:
### Summary
This adds the totals api

### Technical considerations
We need optional states for this to be completed, that's why its currently a draft

I need someone people who are used to Swift to tell me if the Codable and the state initialization bits are done properly

### Relationships
Closes #501
Closes #503
Closes #504
Closes #510
Closes #512
Closes #513
Closes #514

### Pull Request checklist

### Pointers
Go at it commit by commit

### Review template

## Merge permissions
Anyone, anytime


Modules/API/API/Models/GraphItem.swift
```diff
@@ -0,0 +1,17 @@
...
+struct GraphItem: Decodable {
+    let seconds: TimeInterval
+    let byRate: [String: TimeInterval]
+
+    private enum CodingKeys: String, CodingKey {
+        case seconds
+        case byRate = "by_rate"
+    }
+
+    public init(from decoder: Decoder) throws {
```
juanlaube, line 12: I think you don't need this implementation since you are not doing anything special to decode each property, the compiler should be able to synthesize it you 🤔

Modules/API/API/Models/TotalsResponse.swift
```diff
@@ -0,0 +1,23 @@
...
+    let groups: [GraphItem]
+
+    enum CodingKeys: String, CodingKey {
+        case startDate = "start_date"
+        case endDate = "end_date"
+        case resolution
+        case groups
+    }
+
+    public init(from decoder: Decoder) throws {
```
juanlaube, line 16: Same here

Modules/API/API/API.swift
```diff
@@ -6,10 +6,13 @@ import Models
...
     #if DEBUG
-    private let baseURL: String = "https://mobile.track.toggl.space/api/v9/"
+    private static let baseUrl: String = "https://mobile.track.toggl.space"
     #else
-    private let baseURL: String = "https://mobile.track.toggl.com/api/v9/"
+    private static let baseUrl: String = "https://mobile.track.toggl.com"
     #endif
+
+    private let apiUrl: String = "\(baseUrl)/api/v9/"
+    private let reportsUrl: String = "\(baseUrl)/reports/v3/"
```
Odrakir, line 15: You can use `+` to append strings, which is a little cleaner.
```suggestion
    private let apiUrl: String = baseUrl + "/api/v9/"
    private let reportsUrl: String = baseUrl + "/reports/v3/"
```

Modules/Reports/Reports/ReportsState.swift
```diff
@@ -8,17 +8,21 @@ public struct ReportsState: Equatable {
         internal var startDate: Date = Date()
         internal var endDate: Date?
         internal var totals: Loadable<ReportsTotals> = .nothing
+        internal var selectedWorkspace: GenericIdentifier<Workspace> = 0.toIdentifier()
```
Odrakir, line 11: Why this default workspace?

Modules/API/API/API.swift
```diff
@@ -162,3 +175,46 @@ extension API: FeedbackAPI {
...
+        if let endDate = endDate {
+            data["end_date"] = "\(endDate)"
+        }
+        
+        do {
+            let endpoint: Endpoint<TotalsResponse> = try createReportsPostEndpoint(
+                path: "workspace/\(workspaceId.rawValue)/search/date_entries/totals",
+                body: data
+            )
+            return urlSession.load(endpoint).asSingle().map { response in
```
Odrakir, line 201: Are we going to use `TotalsResponse` anywhere outside this at all? Couldn't we map the server response directly into `ReportsTotals` with a custom implemenation of Codable? Those two seem very similar.

heytherewill: My idea here is to let the response free in case we wanna do something else with it and let the actual mapping logic live somewhere else (most likely in an effect)

Modules/API/API/API.swift
heytherewill: Concatenation is only needed in language that messed up their string interpolation syntax 🤭

Modules/Reports/Reports/ReportsState.swift
heytherewill: I would actually rather make this optional and default to the user's workspace

Modules/API/API/Models/GraphItem.swift
heytherewill: Is it able to infer the `snake_case` automagically?

juanlaube: Not automagically, but close enough, you need to set the decoding strategy in the JSONDecoder, check https://developer.apple.com/documentation/foundation/jsondecoder/keydecodingstrategy/convertfromsnakecase

PR #671, opened 2020-08-04
Add the rest of the password reset flow

### LINKED ISSUE #574: Forgot password - add EmailEntered action
## Description
Add the action for handling email input on the forgot password screen. Wire it up to the UI if it’s already available.
## Definition of done
- Action is added and reduced
- It’s wired with the UI if the UI is there
- Tests are added

### LINKED ISSUE #576: Forgot password - add PasswordResetEmailSent and PasswordResetEmailFailed action
## Description
Add the relevant actions. Display an appropriate toast when email is successfully sent. Set an error if the reset email couldn’t be sent.
## Definition of done
- Actions are added and reduced
- They are wired to the UI if the UI is there
- Tests are added

### LINKED ISSUE #577: Forgot password - add SendEmailButtonTapped action, SendPasswordResetEmailEffect and the API interaction
## Description
Add the action and effect. Wire the action up with the UI if it’s there. See the Xamarin implementation for request/response spec.
## Definition of done
- Action is added and reduced
- Effect is added
- API is integrated
- Tests are added

dshr:
## Info
This is adds all the extra stuff for the password reset flow.

I'm not sure about some bits, see comments.

### Relationships
Closes #577
Closes #576
Closes #574

### Pull Request checklist

### Pointers
Commit by commit should work best. For running comment out `LoginViewController.swift:106-108` and move `PasswordResetViewController.swift:38-92` into a separate func called via a `DispatchQueue.main.asyncAfter`

### Review template
- [ ] The code considers i18n

## Merge permissions
This should wait for the routing/optional state refactor first. Merge only after the relevant translation issues have been created.


Modules/Onboarding/Onboarding/PasswordReset/PasswordResetReducer.swift
```diff
@@ -1,14 +1,63 @@
...
+        switch self {
+        case .emptyEmail:
+            return Strings.Validation.emailValidationEmptyError
+        case .noAtSign:
+            return Strings.Validation.emailValidationNoAtSignError
+        case .invalidEmail:
+            return Strings.Validation.emailValidationInvalidEmailError
+        }
+    }
+}
```
dshr, line 22: I'm not too sure that this should live in this reducer, but idk where would be the right place to put this?

Odrakir: We don't have a central place for error handling, I think this is fine here for now.

Modules/Onboarding/Onboarding/PasswordReset/PasswordResetReducer.swift
```diff
@@ -32,3 +49,15 @@ func createPasswordResetReducer(api: UserAPI) -> Reducer<OnboardingState, Passwo
         }
     }
 }
+
+func sendEmailEffect(_ email: String, via api: UserAPI) -> Effect<PasswordResetAction> {
+    guard email.count > 0
+        else { return .just(.passwordResetEmailFailed(EmailValidationError.emptyEmail.toErrorType())) }
+    guard email.contains("@")
+        else { return .just(.passwordResetEmailFailed(EmailValidationError.noAtSign.toErrorType())) }
```
Odrakir, line 57: Do we need a specific error for this? Isn't it caught by the regular expression?

Modules/Onboarding/Onboarding/PasswordReset/Views/PasswordResetViewController.swift
```diff
@@ -28,19 +28,71 @@ public class PasswordResetViewController: UIViewController, Storyboarded {
...
+
+        store.select(\.passwordResetResult)
+            .map { $0 != .inProgress }
+            .drive(submitButton.rx.isEnabled)
+            .disposed(by: disposeBag)
+
+        store.select(\.passwordResetResult)
+            .map { $0 != .inProgress }
+            .drive(closeButton.rx.isEnabled)
+            .disposed(by: disposeBag)
```
Odrakir, line 81: Even though `select` returns a Driver and it's shared between all subscribers, don't you think this would be better in just one chain?

```swift
        store.select(\.passwordResetResult)
            .drive(onNext:handlePasswordResetResult)
            .disposed(by: disposeBag)
```

Modules/Onboarding/Onboarding/PasswordReset/PasswordResetReducer.swift
dshr: It is indeed caught by the regex, but our spec calls for three separate error messages 🤷‍♂️

Modules/Onboarding/Onboarding/PasswordReset/Views/PasswordResetViewController.swift
dshr: Oh, nice! Didn't think of that, will implement it asap

dshr:
@Odrakir the `The store was sent the action while it was already processing another action.` will def be fixed by the new routing. For the time being you can test this by moving `PasswordResetViewController.swift:38-92` into a separate func called via a `DispatchQueue.main.asyncAfter` with a short delay. You could just wrap the whole block into the closure, but then you'd have to add `self` everywhere...

Odrakir:
@dshr Adding this after line 43 on `ResetPassworViewController` fixes it:

```swift
.observeOn(MainScheduler.asyncInstance)
```

PR #676, opened 2020-08-07
Store user preferences

### LINKED ISSUE #439: Make UserPreferences persistable and updatable in the Repository and Database
## Description
Make UserPrefernces persistable and updatable in the Repository and Database.

## Dependencies
#418 #420
## Definition of done
- [ ] Make UserPrefernces persistable and updatable
- [ ] Use this in the UpdateUserPreferences effect
- [ ] Add tests

Odrakir:
## Info
Adds the necessary code to store user preferences in `UserDefaults`.

### Summary
This extends the `UserPreferences` struct to make it `Codable` and then adds another property wrapper to store `Codable` objects in `UserDefaults` 

### Technical considerations
I tried to avoid creating another property wrapper for this and instead using the `UserDefault` one, but I didn't manage to make it work so I could save both plain types and `Codable` ones. Any ideas here are welcome.

Alternatively user preferences could be stored in the DB, but I think this makes more sense in `UserDefaults`. This is done by the `Repository`, which has that dependency, so we can still sync it with the server and it's transparent for the rest of the app.

### Relationships
Closes #439 

### Pull Request checklist

### Pointers
One commit at a time

### Review template

- [ ] The code considers i18n

## Merge permissions
Go!


Modules/OtherServices/OtherServices/MockUserDefaultService.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 
 #if DEBUG
-public class MockUserDefaultService: UserDefaultService {
+public class MockUserDefaultService: UserDefaultsService {
```
dshr, line 4: Shouldn't this then be `MockUserDefaultsService`?

Modules/Settings/Settings/Common/SettingsReducer.swift
```diff
@@ -15,7 +15,7 @@ public func createSettingsReducer(
     calendarService: CalendarService,
     platformInfo: PlatformInfo,
     feedbackAPI: FeedbackAPI,
-    userDefaultsService: UserDefaultService
+    userDefaultsService: UserDefaultsService
```
dshr, line 18: Why are we injecting it here? I thought it all had to go through the Repository?

Modules/TogglTrack/TogglTrack/AnalyticsReducer.swift
```diff
@@ -6,7 +6,7 @@ import Settings
 import OtherServices
 
 func createAnalyticsReducer(analyticsService: AnalyticsService,
-                            userDefaultService: UserDefaultService)
+                            userDefaultService: UserDefaultsService)
```
dshr, line 9: Same thing here

Modules/TogglTrack/TogglTrack/App/AppEnvironment.swift
```diff
@@ -18,7 +18,7 @@ public struct AppEnvironment {
     public let analytics: AnalyticsService
     public let calendarService: CalendarService
     public let notificationsService: NotificationsService
-    public let userDefaultService: UserDefaultService
+    public let userDefaultService: UserDefaultsService
```
dshr, line 21:
```suggestion
public let userDefaultsService: UserDefaultsService
```

Modules/Models/Models/UserPreferences.swift
```diff
@@ -89,8 +89,49 @@ public enum DateFormat: Equatable {
...
+    }
+
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(selectedWorkspaceId?.rawValue, forKey: .selectedWorkspaceId)
+        var smartAlertsStateTimeBefore: Double?
+        if case let .enabled(timeBefore) = smartAlertsState {
+            smartAlertsStateTimeBefore = timeBefore
+        }
+        try container.encode(smartAlertsStateTimeBefore, forKey: .smartAlertsState)
```
sofokli1, line 127: why not put this in the `if` and drop the var altogether?

Modules/Settings/Settings/Common/SettingsReducer.swift
Odrakir: It doesn't necessarily have to always go through the repo, we could, however split it in interfaces to make that difference clearer. I'll take a note.

sofokli1:
Taking this over to fix and merge

PR #680, opened 2020-08-10
Time entry subscriptions

### LINKED ISSUE #620: Implement Time Entry subscription
## Description
Implement subscription that will constantly update time entry list whenever there is a change in the database. 

## Definition of done
- [ ] Refactor repository to provide continuous updates
- [ ] Implement subscription that will provide continuous time entry updates
- [ ] 🔥 Old code that was responsible for time entry updates

Odrakir:
## Info
This replaces the way we update the state after a DB change with subscriptions. Only for TimeEntries for now.

### Summary
Instead of the repository methods returning an Observable, now they'll just throw if there's an error, and updates will come from a direct observation to the DB via a property exposed by the repository called `timeEntriesChanges`.

This will help with sync, since all updates to the DB will come to the state on the same path. On the other hand this makes the repository tests less complete, since we don't wait for specific changes to the state after an action is dispatched, this will be mitigated with integration tests, and also I'll add an issue to write tests for the new Repository obervable property.

### Technical considerations
Not much to say, this works on top of subscriptions, which were added in a previous PR. This subscription will be enabled when the user is logged in and disabled when he's not.

### Relationships
Closes #620 

### Pull Request checklist

### Pointers
One commit at a time

### Review template

- [ ] The code considers i18n

## Merge permissions
Do it!


dshr:
Taking over

Modules/Models/Models/EntityCollection.swift
```diff
@@ -91,6 +91,22 @@ public struct EntityCollection<EntityType>: MutableCollection, RandomAccessColle
...
+        var low = 0
+        var high = self.count - 1
+        while low <= high {
+            let mid = (low + high)/2
+            if isOrderedBefore(self[mid], entity) {
+                low = mid + 1
+            } else if isOrderedBefore(entity, self[mid]) {
+                high = mid - 1
+            } else {
+                self.insert(entity, at: mid)
```
simonrozsival, line 104: I believe there's a missing return here - the code will continue and eventually it will end the `while` loop and then it would insert the entity one more time (line 107). I don't suppose that's the intention.

simonrozsival: I think there should be tests for this method, it's not entirely trivial (inserting into an ordered array using binary search).

Modules/Models/ModelsTests/EntityCollectionTests.swift
```diff
@@ -65,4 +65,40 @@ class EntityCollectionTests: XCTestCase {
...
+        assert(entityStore[4].id == 4)
+        assert(entityStore.count == 5)
+    }
+    
+    func test_insertIsOrderedBefore_whenOrderingIsNotObvious_insertsValuesCorrectly() {
+        let entities = [0, 1, 2, 3].map { TestEntity(id: $0.toIdentifier(), name: "Entity \($0)") }
+        var entityStore = EntityCollection(entities)
+        let insertedEntity = TestEntity(id: 4.toIdentifier(), name: "Entity 4")
+        entityStore.insert(insertedEntity, isOrderedBefore: { $0.name.count < $1.name.count })
+        assert(entityStore[1].id == 4)
```
simonrozsival, line 101: Now that I see this test, the ordering now feels arbitrary - the algorithm won't yield the same results if you insert the same elements in different orders 🤔 I suppose it should be inserted either at the end or at the beginning of all the entities which are in the same equality class. In this case just by using `<=` would lead to such behavior - it would be inserted at the end of the list. Maybe it's not hugely important, because in practice we'll be using this for TEs and we'll be ordering them by the start time, which will be mostly different for all of the TEs and the ordering will be just fine.

simonrozsival: Although to be fair - even with `<=` the ordering would be different if you inserted the elements in a different order. The best thing to do in practice would be to use a comparison such that there is always a linear ordering of elements (e.g., compare `start` and `id` if start times match). I think this function is just fine for now and that we can move forward with this.

PR #700, opened 2020-08-14
Add the terms of service view

### LINKED ISSUE #617: Add Terms and Conditions, Privacy Policy UI
## Description
During third party logins, we can detect there’s no account associated with the email address. We then switch to signup, but before actually signing the user up, we need to show them a screen where t&c + pp can be accepted. Note that we no longer present the country selection list and we have no designs for this screen, so please talk to product about how this should look. Maybe a simple popup would be sufficient, or just what we had before with country selection, but without the actual country list.

## Definition of done

- [ ] UI is added, where t&c + pp is linked and can be accepted

### LINKED ISSUE #644: Add AcceptTermsButtonTapped action
## Description
Add the action, wire it up to the UI. It should be emitted when user accepts terms and conditions. The action emits SignUserUpWithThirdPartyEffect or SignUserUpEffect, depending on where it originated (welcome screen / email, password signup screen).

## Dependencies
#617 
## Definition of done
Action is added, reduced, tested
It’s wired with the UI

### LINKED ISSUE #647: Add MustAcceptTerms action
## Description
Add the action and reduce it so that the terms and conditions view from #617  is shown to the user

This action should be accessible in welcome and signup submodules.

## Dependencies
#617 
## Definition of done
- Action is added, reduced, tested

### LINKED ISSUE #677: Add RejectTerms action
## Description
If user rejects the terms (by closing the UI #617 ) during third party signup, we need to emit this action and reduce it so the welcome screen is shown and user is not signed up.

## Dependencies
#617 

## Definition of done
- [ ] action is added, reduced, tested
- [ ] UI is wired up

### LINKED ISSUE #678: Introduce accept terms screen to the email/password signup flow
## Description
Depending on how the discussion here https://toggl.slack.com/archives/CUHSVKLUV/p1596792481053600?thread_ts=1596013389.005900&cid=CUHSVKLUV goes, we'll have to add some copy to the signup screen, or introduce another step during signup - showing #617 

This issue is here so we don't forget to take action, but is not actionable until decisions are made in the slack thread.

**Update**
Decision has been made to show the modal from #646 during email/pass signup. Therefore, signup has to be changed so that (see the definition of done section):

## Dependencies
#617 
#647 
#644 

## Definition of done
- [ ] tapping signup button emits MustAcceptTerms
- [ ] after accepting terms, signupeffect is emitted
- [ ] when terms are not accepted, user is taken back to the signup screen

dshr:
## Info
This PR adds a reusable terms of service screen
### Relationships
Closes #617
Closes #647
Closes #677
Closes #678 
Closes #644 

### Pull Request checklist

### Pointers
Commit by commit should work best!

### Review template
- [ ] The code considers i18n

## Merge permissions
Go for it!


Modules/Onboarding/Onboarding/Common/OnboardingState.swift
```diff
@@ -23,7 +23,7 @@ public struct OnboardingState: Equatable {
 
         case main
         case emailLogin(showPasswordReset: Bool = false)
-        case emailSignup
+        case emailSignup(showTermsOfService: Bool = false)
```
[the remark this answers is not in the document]
dshr: The ToS screen doesn't depend on the email signup screen at all though. You can easily have a `.thirdPartyLogin(showTermsOfService: Bool = false)` and not have to change a thing in the ToS code.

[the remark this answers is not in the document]
simonrozsival: I see, but then there would be two coordinators which navigate to that screen, that's kinda strange 🤔 but OK, seems like a reasonable explanation.

Modules/Onboarding/Onboarding/TermsOfService/Views/TermsOfServiceViewController.swift
```diff
@@ -0,0 +1,124 @@
...
+            presentWebView(url: privacyPolicyUrl)
+        }
+    }
+    
+    private func presentWebView(url: String) {
+        let webView = SFSafariViewController(url: URL(string: url)!)
+        webView.modalPresentationStyle = .pageSheet
+        webView.preferredBarTintColor = Color.backgroundOnboarding.uiColor
+        webView.preferredControlTintColor = Color.onboardingTextPrimary.uiColor
+        present(webView, animated: true, completion: nil)
```
sofokli1, line 116: shouldn't this happen in the coordinator as we usually do with presentations?

dshr: Nah, these don't contain any state data or emit actions, so no need to wrap them IMO.

sofokli1: shouldn't there be actions `showPrivacy`, `showTos`?

sofokli1: more like `privacyTapped`

dshr: Why? All we're doing here is show a `SFSafariViewController`.

sofokli1: if you think about it you are bypassing the routing approach we have now. The view should not be presenting things on its own without the state knowing about it.

dshr: The routing approach only extends to views that are parts of our app, and not all of them at that. These are `SFSafariViewController`s, so they are not a part of our app.

sofokli1: you are initialising, configuring and presenting the view, that is exactly what we do with the rest of the app. just because we are not subclassing this VC like we do with our UIViewController subclasses doesn't mean it should be used in a different way. We create the instance and our code owns it, this is not an alert view that comes from the OS like when asking for calendar permissions or location data that we do not own in our code. I thought the point of our routing was to see all the navigation code in the coordinators constantly and not look for navigations hidden in VCs.

dshr: The point of our routing is to connect our VCs to the state and reducers in a clean manner. These two are not connected to the state in any way.

sofokli1: SFSafariViewControllers is our VC from the moment we instantiate it. If you do not want to change it is fine (I am not asking for changes, mostly conversing), but this will be come up again when we show other web views for settings and what not. To me it looks really strange that a VC can decide to present modally on its own and change the views stack without any action. What triggered this presentation, no action for that? No action means also no possibility to tests.

Modules/Onboarding/Onboarding/EmailSignup/EmailSignupReducer.swift
```diff
@@ -6,43 +6,76 @@ import API
...
+                return .none
+                
+            case let .setError(error):
+                state.user = .error(error)
+                return .none
+                
+            case .termsOfService(.acceptTermsButtonTapped):
+                state.user = .loading
+                state.screen = .emailSignup(showTermsOfService: false)
+                return signUserUpEffect(email: state.email,
```
simonrozsival, line 46: What will happen when the user uses third party auth to sign up - there's no email or password then. I'm currently thinking about this because I'm (still) working on Google&Apple login, which is closely tight to sign up though these third party services. I believe we can add a property for the type of sign up later and switch that and call either one of them, but it's kinda weird. Wouldn't it be simpler and cleaner to pass an `OnboardingAction` to the ToS screen which would be dispatched as soon as the user taps the CTA button? In this case it would be some action like `.confirmEmailSignup`, in the case of third party sign-up it would be `.confirmThirdPartySignUp(token)` or something like that... What do you think about that?

dshr: But that would be in a whole different reducer/action setup, I don't see how this will impact it? In the `ThirdPartyLoginAction.swift` you'd have a similar `.termsOfService(TermsOfServiceAction` action, and the corresponding reducer will handle a `ThirdPartyLoginAction.termsOfService(.acceptTermsButtonTapped)` action.

simonrozsival: OK, let's not overengineer this 👍

simonrozsival (approved):

sofokli1 (approved):

PR #701, opened 2020-08-17
Calendar: Add calendar header

### LINKED ISSUE #481: Settings: Add settings view
## Description
- Adds the settings button to the calendar and log view (check designs). 
- Decide on how to better construct the view (ideally take inspiration from the modular way it was done in the old app)
- The settings table view with the needed cells:
1. for settings that show the selected value on one side
2. for settings that have a switch
3. For settings that just open other views (like Help)
4. Sign out
- For type 1 and 2 cells values should come from the state.

(Tapping on a cell does nothing for now)

## Dependencies
None
## Definition of done
- [x] Add settings button to calendar tab (or make an issues if not possible) (check designs for correct positioning)
- [x] Add settings button to log view (check designs for correct positioning)
- [x] Chose a logic for the structuring of the settings internally that can be used to populate the table view
- [x] Add the tableview and the cells

### LINKED ISSUE #672: Calendar: Add calendar header
## Description
Add header above the date picker. So add a navigation bar on top with title and the settings button and a view for the totla time tracked and date between the navigation bar and the date picker. Check designs.
## Dependencies
none
## Definition of done
- [x] Add the navigation bar
- [x] Add the settings button
- [ ] Add the labels for the total amount of time tracked in the current day + logic (it should update if there is a running TE)
- [x] Add the date label + logic

sofokli1:
### Summary
Adds the navigation bar to the cal view, setting button, and 2 labels at the top for date and duration.

### Technical considerations
Main issue I am having is that the navigation does not collapse when scrolling because this collection view is nested deeper in a container view in a pager view. I do not want to make any hacky approach by checking scroll position and such in order to fake the default behaviour since there is no way we will mimic it correctly. I am leaning toward leaving an issue and work on this slater when we want to polish this. If anyone wants to suggest anything I am all ears.

Second minor issue is the duration label is not updating when there is a running time entry because the state is not changing. Same problem is visible in the section header of the main log with a running time entry (you have to scroll it out of the visible rect and back to see it update). I was thinking of leaving an other issue for this too since it is related to a hight conversation on how do we want to handle this cases where the state should update every second.

### Relationships
Closes #672
Closes #481

### Pull Request checklist

### Pointers
run the app, commit by commit

### Review template
- [ ] The code considers i18n

## Merge permissions
🦑


Modules/Calendar/Calendar/Common/CalendarReducer.swift
```diff
@@ -5,35 +5,75 @@ import CommonFeatures
...
-                        .subscribe()
-                }
-                return .none
-            case .calendarDay:
-                return .none
-            case .contextualMenu:
-                return .none
-            case .datePicker:
-                return .none
+    let settingsReducer: Reducer<CalendarState, CalendarAction> = createSettingsReducer(
```
simonrozsival, line 25: This doesn't seem like a good idea to me - we're basically duplicating the same code as in `TimerReducer`. There must be a better way of doing this (but maybe not with our current architecture)..

simonrozsival (approved):
It works. It doesn't match the designs 100%, but I don't think that's a big issue at this point of development (@dshr might not agree though 😄).

I don't like that we need to put the duplicate code for opening the settings view in the reducer, but I can't really see a better way of doing this either 🤷‍♂️ Maybe if the settings view would be another of the tabs... but that's not for us developers to decide. Hopefully we'll be able to improve this later on.

dshr (approved):

PR #707, opened 2020-08-20
Force setting's toggles' flipping animation

### LINKED ISSUE #697: Investigate a better way for handling UISwitch animations in settings
## Description
The problem with the switches in settings is that tapping the switch triggers the reducer, which changes the state, which triggers the selector for the settings rows, which reloads the table view before the switch has time to do the animation from on to off (or the other way). Find a better way for allowing the switch to make the animation and update the state (ideally without delaying the change in the value).
Ref #692 where this was introduced

## Dependencies
none

## Definition of done
- [ ] Find a solution that allows the switch to finish its animation (ideally without delaying the change in the value)

simonrozsival:
### Summary
This is a suggestion of a fix for a glitch in the settings view. I'm not 100% sure this is a good permanent fix for the issue, but it works and it's very simple.

### Technical considerations
The problem is caused by the way we subscribe to the store: whenever the state of the store changes, the whole table view which forms the settings view is reloaded (`reloadData()`) and this prevents any kind of animation in the table rows.

I thought of two workarounds:
- wait for the automatic animation of the flipped switch to finish and only then dispatch the action that the value changed (the simplest way to implement this is to add a 100ms delay to the rx stream in `ToggleRowCell` which dispatches the action
- when a table cell is re-configured during a reload, force the animation of the switch flip

I decided to go with the second option for now, because it changes the state first and it animates the change, unlike the first option, which animates the change and shows it to the user and only then it changes the state.

In the end, I don't think this is an optimal solution for the following reasons:
- the `ToggleRowCell` is now tightly bound to the table view's behavior (`RxTableViewSectionedReloadDataSource`) and if we changed the behavior of the view source (in a way such that the whole table is not reloaded anymore), the switches could start behaving in a weird way (flipping back and forth or something like that)
- the settings view is very short and I believe iOS doesn't bother reusing the cells and so everything looks nice. If the behavior of the table view changed for some reason (if it started reusing cells more aggressively), the animations might not work every time 🤷‍♂️ 
- It's a hack, not a good solution. The decision to reload the settings table every time probably saved some time during development (there are no bindings between state properties and individual settings items), but it doesn't seem very robust to me. I would prefer changing the settings view into a `UIStackView` (generated dynamically from the same list of VMs), but that would require significant refactoring, which is probably not a good idea at the moment (we need devs to focus on adding features, not refactoring stuff that works)

### Relationships
Closes #697 

### Pull Request checklist
- [ ] ~The code in this PR is tested thoroughly enough that changing the behavior to an incorrect one breaks tests~

### Review template

- [ ] The code considers i18n

## Merge permissions
🤷‍♂️ tell me what you think and let me know if you have a better idea to solve this issue


dshr (approved):
LGTM

sofokli1 (approved):
good enough for now

Issue #492, opened 2020-06-29
Settings: Implement Smart Alerts setting


sofokli1:
## Description
I suggest using the `Settings: Implement select from list sub-setting` (check dependencies below for a link) to show the selectable options. And to use the type 1 cells created in `Settings: Add settings view` issue. The view should take into account the notification permissions since based on that it should show a button (to open the os settings to enable notifications) or the list of options.

## Dependencies
`Settings: Add settings view` #481 
`Settings: Implement select from list sub-setting`: #486
## Definition of done
- [ ] Implement the logic
- [ ] implement the UI

sofokli1:
waiting for new behaviour from design and feature team

PR #711, opened 2020-08-27
Add more settings

### LINKED ISSUE #490: Settings: Implement Workspace selection
## Description
Present the options using what was implemented in `Settings: Implement select from list sub-setting`. Check dependencies for links.

## Dependencies
`Settings: Implement select from list sub-setting`: #486
## Definition of done
- [x] Implement the logic
- [x] implement the UI

### LINKED ISSUE #494: Settings: Implement About
## Description
Should include the same options as the old app
- Privacy policy
- Terms of Service
- Licenses

## Dependencies
## Definition of done
- [ ] Implement the logic
- [ ] implement the UI

### LINKED ISSUE #495: Settings: Implement Help
## Description
Same as the old app.

## Dependencies
## Definition of done
- [ ] Implement the logic
- [ ] implement the UI

sofokli1:
### Summary
The smart alert part. is not complete and it is on hold until the feature team agrees on what will happen with it. There is talk of changing it and also move it in the cal settings view. The rest of the issues should work correctly.

### Relationships
Closes #495
Closes #490
Ref #492
Closes #494

### Pull Request checklist

### Pointers
Go commit by commit

### Review template
- [ ] The code considers i18n

## Merge permissions
🦑


Modules/Assets/Assets/Base.lproj/Localizable.strings
```diff
@@ -114,3 +114,12 @@
...
 "nameValidationInvalidNameError" = "That name doesn't look right";
+"defaultWorkspace" = "Default Workspace";
+"upcomingEvent" = "Upcoming Event";
+"disabled" = "Disabled";
+"whenEventStarts" = "When event starts";
+"minutesBefore" = "%d minutes before";
+"oneHourBefore" = "1 hour before";
+"Privacy Policy" = "Privacy Policy";
+"Terms of Service" = "Terms of Service";
+"Licenses" = "Licenses";
```
Odrakir, line 125: Are there issues to translate these?

Modules/Settings/Settings/Common/SettingsSelector.swift
```diff
@@ -76,10 +83,27 @@ let multiOptionSettingsSections: (SettingsState) -> [SettingsSectionViewModel] =
...
+        }
+        sections.append(SettingsSectionViewModel(original: .subSettingSelection([]), items: rows))
+    case .about:
+        let rows = [
+            NavigationRow.init(title: Strings.Settings.privacyPolicy, action: .openLink(URL(string: privacyPolicyUrl)!), detail: nil),
+            NavigationRow.init(title: Strings.Settings.termsOfService, action: .openLink(URL(string: termsOfServiceUrl)!), detail: nil),
+            NavigationRow.init(title: Strings.Settings.licenses, action: .openLicenses, detail: nil)
+        ]
+        sections.append(SettingsSectionViewModel(original: .subSettingSelection([]), items: rows))
     case .none, .some(.none):
```
Odrakir, line 103: There's a warning here, we should be using `nil` instead of `.none` in the first case.

Modules/Settings/Settings/Common/SettingsState.swift
```diff
@@ -60,5 +60,6 @@ public enum MultiOptionSetting: Equatable {
     case firstDayOfTheWeek(DayOfTheWeek)
     case dateFormat(DateFormat)
     case durationFormat(DurationFormat)
+    case defaultWorkspace(Workspace.Identifier?)
```
Odrakir, line 81: Can this ever be nil? 🤔

Modules/Models/Models/UserPreferences.swift
```diff
@@ -71,6 +66,42 @@ extension UserPreferences: CustomStringConvertible {
     }
 }
 
+public enum SmartAlertOption: Equatable {
```
Odrakir, line 69: I think I know where you are coming from, but to me the name was fine as `SmartAlertState`. Then you can have an `allOptions` property instead of `allCases`. What do you think?

Podfile
```diff
@@ -236,4 +236,7 @@ post_install do |pi|
         config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'] = '12.0'
       end
     end
+
+    # The following line converts the acknowledgements.markdown to html to show in the the lincens view in app settings
```
Odrakir, line 240: Are you sure this should be in the Podfile? We might ditch Cocoapods at some point losing this and probably not realizing about it.

Odrakir (changes requested):

Modules/Models/Models/UserPreferences.swift
sofokli1: I find `SmartAlertState` confusing because of the way we use the word `State` elsewhere. It does not look consistent.
As for `allOptions` I am ok with that change

Podfile
sofokli1: If and when we drop cocopods we should think of an alternative, until then this is fine. This the simplest way for now. If you have an alternative I am all ears.

Modules/Assets/Assets/Base.lproj/Localizable.strings
sofokli1: ref #715

Modules/Settings/Settings/Common/SettingsSelector.swift
sofokli1: without `.none, .some(.none)` xcode complains that switch is not exhaustive

sofokli1: I see not warning. are using the latests xcode and swiftlinter?

Modules/Settings/Settings/Common/SettingsState.swift
sofokli1: since the `preferences.selectedWorkspaceId` is optional. In `SettingsReducer` line 88 we have the need for the optional associated value. Should I `guard` for nil `preferences.selectedWorkspaceId` before setting `state.multiOptionSetting` in `SettingsReducer`?

Modules/Settings/Settings/Common/SettingsSelector.swift
Odrakir: It happens in Xcode 12, you can fix it by writing the case like this: `case nil, .some(.none):`

Modules/Settings/Settings/Common/SettingsState.swift
Odrakir: Yeah, I think it makes sense to guard agains that, if there's no default workspace we show an alert and the user should never get this far.

Podfile
Odrakir: I'm saying we won't ever remember that's there. Why are we doing this on compile time anyway? Why don't we just embed the html?

Odrakir (changes requested):
Check my answers

Modules/Settings/Settings/Common/SettingsSelector.swift
sofokli1: Will try this when I get home but the nil case may be something new not available in xcode 11

Podfile
sofokli1: This is not compile time. Only when running pod install and similar commands. Which is the right time since that is when the new pods and their licenses are added. We went with this approach because no one will remember to update the licenses manually.

Modules/Settings/Settings/Common/SettingsSelector.swift
Odrakir: It is, I tried.

Podfile
Odrakir: Yeah, not compile time, my bad.

Modules/Settings/Settings/Common/SettingsState.swift
sofokli1: I added the guard but since there is no value, the ws setting will crash the app. Will leave it like this until we are done with the sync and stuff.

Odrakir (approved):

Issue #715, opened 2020-08-31
Add translations 🚫🦑


sofokli1:
## Description
PRs that add new strings should ref this issues in a multiline comment in `Base.lproj/Localizable.strings` with the new strings
Do not close this issue until aurora is released for the first time.

## Definition of done
- [ ] All translations before the first release are added

PR #719, opened 2020-09-02
Add LinkSsoEffect

### LINKED ISSUE #638: Add LinkSsoEffect
## Description
This effect should perform the account linking. See Xamarin app for endpoint spec. Some info is here https://github.com/toggl/mobileapp/pull/7519/files and https://www.notion.so/toggl/SSO-d8929cc996084f3db6fdd08637080f1c

## Definition of done
Effect is added
API is integrated
Results are correctly propagated

sofokli1:
### Summary
Mainly adds the request and the effect to call the request. creates a couple of actions to handle the response of the request. And some tests.

### Relationships
Closes #638

### Pull Request checklist

### Pointers
Commit by commit

### Review template
- [ ] The code considers i18n

## Merge permissions
🦑


Modules/Assets/Assets/Base.lproj/Localizable.strings
```diff
@@ -117,3 +117,5 @@
 "linkAccountTitle" = "An account with this email already exists.";
 "linkAccountSubTitle" = "To enable log in with SSO for this account, please log in with your existing credentials.";
 "linkAccountCTA" = "Log in to enable SSO";
+"badSSOEmailException" = "User email doesn't match SSO credentials";
+"invalidConfirmationCodeSSOException" = "Invalid or expired confirmation code";
```
sofokli1, line 130: ref #715

Modules/Networking/Networking/URLSessionExtension.swift
```diff
@@ -44,7 +44,12 @@ extension URLSession: URLSessionProtocol {
                         observer.onError(error)
                     }                    
                 } else {
-                    observer.onError(NetworkingError.wrongStatus(response.statusCode, response))
+                    do {
+                        let result = try endpoint.parse(data)
```
Odrakir, line 48: I'm not sure if we should be trying to parse the data if there was an error. Why not just return the raw data?

Modules/Onboarding/Onboarding/SSOLink/SSOLinkReducer.swift
```diff
@@ -0,0 +1,34 @@
...
+import Repository
+import OtherServices
+import API
+
+func createSSOLinkReducer() -> Reducer<OnboardingState, SSOLinkAction> {
+    return Reducer {state, action in
+
+        switch action {
+
+        case .setSSOLinkResult:
```
Odrakir, line 14: What's with the string? We don't need it?

Odrakir (changes requested):

Modules/Networking/Networking/URLSessionExtension.swift
sofokli1: that is how it is done in the old app https://github.com/toggl/mobileapp/pull/7519/files#r439433938

Odrakir: It's doing it right before doing the string comparison, which is what I'm saying.

Modules/Onboarding/Onboarding/SSOLink/SSOLinkReducer.swift
sofokli1: no we don't, will fix

Modules/Networking/Networking/URLSessionExtension.swift
sofokli1: ok I see, will return the raw data. was thinking also to remove the `Int` from `wrongStatus(Int, HTTPURLResponse?, Any?)` since it is part of the response anyways. what do you think?

Modules/API/API/API.swift
```diff
@@ -205,6 +205,30 @@ extension API: UserAPI {
     public func removeAuth() {
         headers["Authorization"] = nil
     }
+
+    public func linkSSO(email: String, confirmationCode: String) -> Single<String> {
+        let endpoint: Endpoint<String> = createEntityEndpoint(path: String("me/enable_sso?email=\(email)&confirmation_code=\(confirmationCode)"))
```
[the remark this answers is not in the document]
sofokli1: will take a deeper look at this. first time touching aurora network code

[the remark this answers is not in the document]
sofokli1: Should I do any type of encoding to the email value before putting git in the query dict?

Modules/API/APITests/UserAPITests.swift
```diff
@@ -135,7 +135,9 @@ class UserAPITests: XCTestCase {
         _ = api.linkSSO(email: "some@email.com", confirmationCode: "someComfirmationCode")
 
         XCTAssertEqual("GET", mockURLSession.httpMethod)
-        XCTAssertEqual(loginURL + "/enable_sso?email=some@email.com&confirmation_code=someComfirmationCode", mockURLSession.requestUrl, "uses correct endpoint")
+
+        XCTAssert((loginURL + "/enable_sso?email=some@email.com&confirmation_code=someComfirmationCode") == mockURLSession.requestUrl
```
sofokli1, line 139: I am open to suggestion to better do write this

Modules/Onboarding/Onboarding/SSOLink/SSOLinkReducer.swift
Odrakir: But then why is the endpoint returning a string? I think you have to create a Endpoint<Void> instead

Modules/Networking/Networking/URLSessionExtension.swift
Odrakir: You are right about removing the Int, yeah

Odrakir (approved):

PR #730, opened 2020-09-08
🔥 coordinators

### LINKED ISSUE #722: Remove Coordinators
## Description
As discussed yesterday, we are going to :fire: coordinators and rely on VCs lifecycle intead

## Dependencies
#720

## Definition of done
- [ ] Coordinators are no more

Odrakir:
## Info
Removes coordinators once and for all. 

### Summary
For most of them is just about removing the file and accessing the VC directly, for some which included some logic I moved that logic to the root VC. A couple of them had more complex stuff (like a nav or tab controller) so I convert them into VCs rather than delete them.

Some clean up could be done, but I'm not sure if it's worth it if we are going to make the move to SwiftUI, let's just leave it for now.

### Relationships
Closes #722 

### Pull Request checklist

### Pointers
Commit history is tidy

### Review template

- [ ] The code considers i18n

## Merge permissions
Go!


sofokli1 (approved):
Looks good

PR #731, opened 2020-09-09
Re enable signin

### LINKED ISSUE #725: Re-enable SSO code
## Description
Everything related to SSO was disabled during the migration to SPM. This is because it depends on some Pods, and those should be imported only in App and not in any other package.

To accomplish this we should abstract them the same way we are doing with `AnalyticsService`.

## Dependencies
#724 

## Definition of done
- [ ] SSO Code is abstracted behind an interface so dependencies are only needed in the main App project
- [ ] The whole SSO behavior is back on working as it was before

Odrakir:
## Info
Actually, contrary to what the issue says this was about re-enabling third party sign in and not SSO.

### Summary
I created another service `SignInService` and injected Google and Apple sign in as wrapped classes in it. Much like we do it for analytics. Then I moved the logic out of the VC and into the reducer.

This whole deal is so we don't need to import the Pods anywhere but in the main app. This is only really needed for Google, but it's a good idea to do it for Apple as well and keep the code consistent.

I think this is the best way, we remove logic from VCs and keep everything encapsulated.

### Relationships
Closes #725 

### Pull Request checklist

### Pointers
Commit history is tidy

### Review template

- [ ] The code considers i18n

## Merge permissions
Go!


Dependencies/TogglTrack/Sources/TogglTrack/TogglTrack.swift
```diff
@@ -9,15 +9,19 @@ import Networking
...
-    public init(window: UIWindow, analyticsServices: [AnalyticsService]) {
+    public init(window: UIWindow, analyticsServices: [AnalyticsService], signInServices: [ThirdPartySignInService]) {
 
-//        let appEnvironment = AppEnvironment.live(analyticsServices: analyticsServices)
-        let appEnvironment = AppEnvironment.local
+        let appEnvironment = AppEnvironment.live(
+            analyticsServices: analyticsServices,
+            signInServices: signInServices
+        )
+//        let appEnvironment = AppEnvironment.local
```
sofokli1, line 24: 🔥

sofokli1 (approved):
overall looks good. there is not way to actually test this until the third-party login is in working condition

Dependencies/TogglTrack/Sources/TogglTrack/TogglTrack.swift
Odrakir: Let's leave that for now so it's easy to switch to local for testing.

PR #733, opened 2020-09-10
Refactor database for transactions

Odrakir:
## Info
This PR makes some changes to the DB so it's easy to perform transactions. It also removes the dependency on Rx from the entity databases, moving that to the main Database class.

Finally I refactor the Repository to make use of the new Database API

### Technical considerations
This is basically about performing multiple DB operations tied to a single background context, for that I had to move some code around.

There's also a semaphore in place so we can't do two writing operations at the same time. Reading should be fine.

### Relationships
Closes #136 

### Pull Request checklist

### Pointers
Commit history is simple, ask me any doubts about Core Data code.

### Review template

- [ ] The code considers i18n

## Merge permissions
Go! 🦑


Dependencies/Database/Tests/DatabaseTests/TransactionTests.swift
```diff
@@ -0,0 +1,169 @@
...
+        super.setUp()
+        database = Database(persistentContainer: mockContainer)
+    }
+
+    open override func tearDown() {
+        flushData()
+        super.tearDown()
+    }
+
+    func testMultipleOperations() {
```
sofokli1, line 46: the format we agreed to follow but not a big deal if you don't feel like changing

Dependencies/Database/Tests/DatabaseTests/TransactionTests.swift
```diff
@@ -0,0 +1,169 @@
...
+    func testMultipleOperationsFail() {
+
+        let workspace = Workspace(id: 0.toIdentifier(), name: "Workspace 0", admin: true)
+        let timeEntry = TimeEntry(id: 0.toIdentifier(), description: "TE", start: Date(), duration: nil, billable: false, workspaceId: workspace.id)
+
+        do {
+            try database.performInBackground  { db in
+                //                try db.timeEntries.insert(entity: timeEntry)
+                do {
+                    // try db.workspaces.insert(entity: workspace)
```
sofokli1, line 73: 🔥,  other one above

Dependencies/Database/Tests/DatabaseTests/TransactionTests.swift
```diff
@@ -0,0 +1,169 @@
...
+                    try db.timeEntries.insert(entity: timeEntry)
+                } catch {
+                    print("in", error)
```
sofokli1, line 76: is this needed?

Dependencies/Database/Tests/DatabaseTests/TransactionTests.swift
```diff
@@ -0,0 +1,169 @@
...
+                    throw error
+                }
+            }
+        } catch {
+            print("out", error)
```
sofokli1, line 81: same question?

sofokli1 (changes requested):
Overall looks good, just a few points

Dependencies/Database/Tests/DatabaseTests/TransactionTests.swift
```diff
@@ -0,0 +1,169 @@
...
+
+    lazy var mockContainer: NSPersistentContainer = {
+        guard let modelURL = Bundle.module.url(forResource: self.model, withExtension: "momd") else { fatalError() }
+        guard let managedObjectModel =  NSManagedObjectModel(contentsOf: modelURL) else { fatalError() }
+
+        let container = NSPersistentContainer(name: model, managedObjectModel: managedObjectModel)
+
+        let description = NSPersistentStoreDescription()
+        description.type = NSInMemoryStoreType
+        container.persistentStoreDescriptions = [description]
```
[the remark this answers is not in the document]
Odrakir: I'm not sure I understand. What are we missing?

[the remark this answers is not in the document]
sofokli1: what the blog post is saying is that if we have any `cascading deletes` it is better to use
```
let description = NSPersistentStoreDescription()
description.url = URL(fileURLWithPath: "/dev/null")
container.persistentStoreDescriptions = [description]
```
since the in `NSInMemoryStoreType` does not support `cascading deletes`.

> when using in-memory store, you will find that certain features like cascading deletes might not work as expected. These features appear to rely on features from the default storage (SQLite) that are not available to an in-memory store.

sofokli1 (approved):
other than this comment https://github.com/toggl/ios/pull/733#discussion_r486969861 it looks good

Dependencies/Database/Tests/DatabaseTests/TransactionTests.swift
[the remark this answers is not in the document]
Odrakir: I'll take a note to look into this in the future. Right now I'm not even sure if we should have cascading deletes at all.

PR #735, opened 2020-09-11
Add all entity subscriptions

Odrakir:
## Info
We already had them for TimeEntries, this PR adds DB subscriptions for the rest of the entities.

### Summary
Adds subscriptions and handles them in `LoadingReducer`.

Then it modifies the repository calls to not return the created or modified entity directly (except when absolutely needed as in createTag)

Finally it changes the initial load methods in the repo to go through the subscription system as well.

The alternative is not doing subscriptions at all. Honestly I'm not 100% sure of the best approach. This seems a bit cleaner when it comes to the sync, probably, but I find two main disadvantages:
- Reducer tests are less useful as they don't update the state anymore
- For some calls like `createTag` we still need to return the newly created Tag so we end up having two paths through which state could be updated.

### Relationships
Closes #595 

### Pull Request checklist

### Pointers
Commit history is tidy

### Review template

- [ ] The code considers i18n

## Merge permissions
Go!


dshr (approved):
Looks good to me! I'll have to rewrite #740 after we merge this though

dshr (approved):

PR #738, opened 2020-09-11
Add more SSO flow parts

### LINKED ISSUE #631: Add AuthenticateWithSso action
## Description
This is just a proxy to emit PerformSsoAuthFlowEffect and pass the sso url.

## Dependencies
#629 
## Definition of done
- Action is added, reduced, tested

### LINKED ISSUE #633: Add PerformSsoAuthFlowEffect
## Description
Navigate user to the URL returned by GetSsoUrlEffect. Use https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession or https://developers.google.com/web/android/custom-tabs/implementation-guide Then either emit SetError / SetUser (if the flow returned an api token) / MustLinkSsoAccounts(email, confirmationCode) - if the account needs to be sso linked. More info on params and how the flow can go are in https://www.notion.so/toggl/SSO-d8929cc996084f3db6fdd08637080f1c and in the Xamarin app

## Dependencies
#631 
## Definition of done
Effect is added
It triggers the auth flow 
It handles the flow so that errors are handled and params extracted from the togglauth:// deep link
It emits the correct actions
Tests are added

### LINKED ISSUE #634: Add MustLinkSsoAccounts action
## Description
This action should modify the state so that the LinkSsoAccounts view is shown

## Dependencies
#633 
#618 
## Definition of done
- Action is added, reduced, tested
- It's emitted from the auth flow effect 

### LINKED ISSUE #637: Add LinkSsoAction
## Description
Add the action, this is a proxy that calls LinkSsoEffect

## Definition of done
Action is added, reduced, tested

sofokli1:
### Summary
- Adds the `ASWebAuthenticationSession` to the `LoginWithSSOViewController`
- Adds a bunch of actions and updates the reducer
- Adds tests

### Technical considerations
The sso logic is a bit elusive and hard to follow. This is not finished but id does close many issues. This should be completed when we can actually talk with the api.

### Relationships
Closes #631
Closes #633
Closes #634
Closes #637

### Pull Request checklist

### Pointers
This does not work so there isn't much to test from a practical point of view. Just take a look at the code

### Review template
- [ ] The code considers i18n

## Merge permissions
🦑


Odrakir:
@sofokli1 I wonder if you considered doing something like what I did for Google and Apple sign in, moving some of the code that you added to `LoginWithSSOViewController` to a service so we can handle all logic from the reducer. What do you think?

sofokli1:
@Odrakir that makes sense but I would leave that for a separate issue since this PR is not trying to finalise the SSO flow and more changes will be needed anyway (as soon as the api starts working).

Odrakir:
I have to disagree, if you think it's a good idea to do it that way it should be done now, even if it's just with placeholder code. Otherwise the next person that continues with this (it might not be you) is going to see that logic in the VC and keep adding stuff there.

sofokli1:
This will be touched a week later. Make an issue and assign me directly. Otherwise this will linger for more than a week affecting all the speed metrics for no good reason.

Odrakir:
I do believe there's a good reason to move that code out of the VC, mostly because it'll be easier to test if it's in services and reducers. And secondly because it'll help with bringing it to other platforms as well.

sofokli1:
My point was to not leave the PR open just for this change while we worked on the mnm week. I would have made the change in a different PR after the mnm week. So this PR stayed open for no good reason (since I was going to fix it anyway separately) and this conversation is unless since the time has passed already and the change might as well happen here now. At no point there was a disagreement that the change is needed in general and that it should happen soon rather than in an unknown time in the future.

Odrakir:
Understood, well, I didn't actually review yet and the week has passed anyway, so... let's get to it now! 🚀

Dependencies/SignInService/Sources/SignInService/SSOService.swift
```diff
@@ -0,0 +1,59 @@
+import Foundation
+import AuthenticationServices
+import RxSwift
+
+public enum SSOError: Error {
+    case generalError
+}
+
+public struct SSOPresentationContextProvider: Equatable {
```
Odrakir, line 9: Why do we need this wrapper?

Dependencies/Onboarding/Sources/Onboarding/SSO/SSOReducer.swift
```diff
@@ -39,17 +43,73 @@ func createSSOReducer(api: UserAPI) -> Reducer<SSOState, SSOAction> {
             state.ssoLoginResult = .inProgress
             return .none
 
-        case .authenticateWithSSO:
+        case let .authenticateWithSSO(ssoURL):
+            return authenticateWithSSOEffect(url: ssoURL, presentation: state.presentationContextProvider, ssoService: ssoService)
+
+        case let .linkSSOAction(email, confirmationCode):
```
Odrakir, line 49: Where's this one dispatched from?

Dependencies/Onboarding/Sources/Onboarding/SSO/SSOReducer.swift
```diff
@@ -17,8 +19,10 @@ func createSSOReducer(api: UserAPI) -> Reducer<SSOState, SSOAction> {
             state.ssoLoginResult = .inProgress
             return getSsoUrlEffect(email: state.email, via: api)
             
-        case .mustLinkSSOAccounts:
-            state.ssoLoginResult = .done
+        case let .mustLinkSSOAccounts(email, confirmationCode):
+            state.email = email
+            state.confirmationCode = confirmationCode
```
Odrakir, line 24: I don't think you are using this `confirmationCode`anywhere.

Odrakir (dismissed):
Can you also rename `getSsoUrlEffect` to `getSSOUrlEffect` for consistency?

Dependencies/SignInService/Sources/SignInService/SSOService.swift
sofokli1: I need to store this in the state and ASWebAuthenticationPresentationContextProviding is a protocol thus not equitable. didn't want to add `==` func to the state.

Dependencies/Onboarding/Sources/Onboarding/SSO/SSOReducer.swift
sofokli1: I do not know. Calling this action was not part of the issues. I think it is called from user action from UI.

Dependencies/Onboarding/Sources/Onboarding/SSO/SSOReducer.swift
sofokli1: I am not using it but on the other hand it is not clear what I should do after I get it there either. It looks like it is needed in linkSSOAction. In general the flow is not complete by far and it is not the purpose of the PR.

sofokli1: as mentioned in slack we keep the confirmationCode in the state https://toggl.slack.com/archives/CD40QCV25/p1600761133017000

Dependencies/Onboarding/Sources/Onboarding/SSO/SSOReducer.swift
sofokli1: https://toggl.slack.com/archives/CD40QCV25/p1600760715016900

Dependencies/SignInService/Sources/SignInService/SSOService.swift
Odrakir: 🤔 It actually might not make much sense for that to be in the state at all. Have you seen how it's done for the `SignInService`? It's injected into the VC that's going to use it and configured from there instead of doing it with an action.

sofokli1: @Odrakir Addressed

Odrakir (approved):

PR #741, opened 2020-09-16
Fix local db generation

Odrakir:
## Info
There were some problems with the changes to the `LocalDataRepository`.

### Summary
Fixes the way we get entities and insert them into the DB. Now they'll go in order and wait for the previous one to finish.

I also added some constants to set the max number of entries and the max number of entries per day (a range)

### Relationships
Doesn't close any issue as this came up during the cross-platform testing

### Pull Request checklist

### Pointers
Just one file. And it's only for debug.

### Review template

- [ ] The code considers i18n

## Merge permissions
Go!


Dependencies/Repository/Sources/Repository/LocalDataRepository.swift
```diff
@@ -1,12 +1,18 @@
+#if DEBUG
```
dshr, line 1: Can't we wrap only the contents of `LocalDataRepository` in an `#if DEBUG`, not the whole file?

dshr (changes requested):
I think the over-eager debug wrapper is breaking non-debug builds.

Dependencies/Repository/Sources/Repository/LocalDataRepository.swift
Odrakir: Yeah, we are using it in release builds too (when we shouldn't). But that's a fix for another time. I do what you say.

Dependencies/Repository/Sources/Repository/LocalDataRepository.swift
```diff
@@ -16,69 +21,92 @@ public class LocalDataRepository: Repository {
...
 
                     let timeEntries = try db.timeEntries.getAll()
                     if timeEntries.count == 0 {
                         self.api.loadEntries()
-                            .observeOn(MainScheduler.asyncInstance)
                             .subscribe(onSuccess: {
-                                try! db.timeEntries.insert(entities: $0)
+
+                                var entriesToInsert = $0.grouped(by: { $0.start.ignoreTimeComponents() })
+                                    .map { $0.dropLast($0.count - Int.random(in: self.maxEntriesPerday)) }
```
dshr, line 102: I'm getting a`Thread 26: Fatal error: Can't drop a negative number of elements from a collection` error here. Do we really need this limiting here, since the textfiles probably already have a sensible number of TEs per day?

dshr (changes requested):

Dependencies/Repository/Sources/Repository/LocalDataRepository.swift
Odrakir: Nah, the files have a bunch of TEs per day, it looks very ugly. I added a check so the random error that happened to you can't happen any more.

dshr (approved):
Works well now, but tests are failing weirdly

PR #742, opened 2020-09-21
Refactor DB<->Repo<->Logic mappings

### LINKED ISSUE #681: Create subscriptions for the all entities
## Description
After #680 is merged we should do the same for the rest of entities, the code should be very similar.

Basically creating an observable  in the Repository for every entity, bind it to DB observation and modify the Repository API to throw instead of returning an Observable.

Optionally we could also do the first load of the state via a subscription, but we could also do this further down the road in another issue.

## Dependencies
#680 

## Definition of done
- [ ] Create subscriptions for Project
- [ ] Create subscriptions for Client
- [ ] Create subscriptions for Workspace
- [ ] Create subscriptions for Tag
- [ ] Create subscriptions for Task

### LINKED ISSUE #683: Move business models mappings to repositories
## Description
Right now we have two types of models in the app:
- DB models (`ManagedTimeEntry`) which are classes that inherit from `ManagedObject` and don't leave the DB Module
- Business models (`TimeEntry`) which are the ones used in the rest of the app including the state and the Repository

The mapping from DB models to business models is happening in the DB module. The problem with that is we need to add a lot of new properties to the models so the Repository can do syncing properly. But we don't want those properties leaking to the rest of the app.

To do this we have the following alternatives:
- Use the `ManagedTimeEntry` models in the repo and map them when moving them between the repo and the rest of the app. I don't like this option because we would expose CoreData.
- Add an additional type of model (`RepoTimeEntry`?) that has all the properties needed in the repo (most if not all of them will match the ones in the DB) and map these models to the existing business models (`TimeEntry`) when leaving the repo.
- The same as above but using an interface for the business model that only exposes the properties we need in the rest of the app (no syncing specific properties will leak)

I think number 3 will probably be the best option, but whoever takes this issue should give it a thought and decide.

WARNING: This issue is not about adding those sync only properties, it's about getting the model separation ready for it.

## Dependencies
It can probably be done before #681 , but in any case it'll requiere less conflict solving if it's done after that one.

## Definition of done
- [ ] We define and map models so there's a clear separation of them between DB <-> Repository <-> State

Odrakir:
## Info
This PR is huge, but do not be afraid, most of it is just brainless renaming of properties everywhere.

The gist of it is we want to have 3 types of models in the app:
- `ManagedTimeEntry`: These are the ones stored in the DB and they won't leave the `Database` module. They'll include some properties we don't need in the UI (regarding syncing, not implemented yet)
- `TimeEntryDAO`: These are the ones the `Repository` the `API` and sync (when we implement it) deal with. They'll include the same properties as `ManagedTimeEntry` but don't depend on Core Data.
- `TimeEntry`: These are the ones that the reducers, state and UI know about. They just have the properties we need. 

Additionally there are some `WhateverDTO` that help send data around. Those need some work, but that's for another PR (in particular StartTimeEntryDTO and CreateTimeEntryDTO, we'll discuss it some other time)

This PR basically moves stuff around to accomplish this, while also trying to keep modules isolated and preventing the possibility of creating models outside were they are supposed to be created.

### Technical considerations
Some things to consider:
- I split the `Models` module in 3: `GenericModels`, `StoreModels` and `Models`. This helps with isolation since while I  want to access `StoreModels` from some services (API, Repo, DB...), I don't want to access them from everywhere where the normal Models are accessible.
- I share the Identifier between normal models and DAOs. Meaning `TimeEntry`'s identifier is `GenericIdentifier<TimeEntryDAO>`. This makes it much easier (and safer) to map between those two, while keeping type safe identifiers.
- I removed the `LocalDataRepository` since it makes no sense anymore. Be sure to have some data in the DB before trying this PR or add it manually.
- I was under the impression the `Repository` would contain the syncing code, but judging by Android's code it's not like that. The sync will have access to the API and Repo. That's something to be aware of to understand some of these changes.

### Relationships
Closes #683 

### Pull Request checklist

### Pointers
Not sure if the commit history will help much since I did and the undid some things. You can look at specific commits to get an isolated view of a specific change, though.

### Review template

- [ ] The code considers i18n

## Merge permissions
🚫 Wait for me


Dependencies/Repository/Sources/Repository/Repository.swift
```diff
@@ -23,70 +25,72 @@ public protocol TimeLogRepository {
...
+    internal init(database: Database, userDefaults: UserDefaultsService, time: Time) {
         self.time = time
         self.userDefaults = userDefaults
         self.database = database
     }
 
-    public static let live = Repository(api: API.live, database: Database(), userDefaults: UserDefaultsService(), time: Time())
-
-    public static let local = LocalDataRepository(api: API.local, database: Database(), userDefaults: UserDefaultsService(), time: Time())
+    public static let live = Repository(database: Database(), userDefaults: UserDefaultsService(), time: Time())
```
dshr, line 93: Does it still make sense to call this singleton `live`? 
```suggestion
    public static let instance = Repository(database: Database(), userDefaults: UserDefaultsService(), time: Time())
```

dshr (approved):
All good, save for one 💄 question.

Dependencies/Repository/Sources/Repository/Repository.swift
Odrakir: My guess is we'll end up having mock ones. Let's leave it for now.

sofokli1 (approved):
Looks good

sofokli1:
Although `StoreModel` could be called `DAOModel` because it was not obvious at a glance

PR #747, opened 2020-09-23
Fix subscriptions firing like crazy.

### LINKED ISSUE #746: DB subscriptions are firing more that they should
## Description
They fire even if there are no changes to the specific object they are listening to (but there are changes to others)

## Definition of done
- [ ] DB subscriptions fire only when a change to the entity they are listening to happens

Odrakir:
## Info
Fixes a small bug and adds a test.

We should probably add more tests to DB subscriptions, but that's for another PR. I'll take a note.

### Technical considerations
I had to move the `observeOn` to the repo in order to test this properly.

### Relationships
Closes #746 

### Pull Request checklist

### Pointers
It's pretty easy stuff

### Review template

- [ ] The code considers i18n

## Merge permissions
Go!


sofokli1 (approved):
looks good and makes sense

PR #770, opened 2020-09-24
Make Workspace syncable

### LINKED ISSUE #769: Make Workspaces syncable
## Description
Initially we were going to do TEs first, but after talking with Will I decided to change plans and do Workspaces first, since it's simpler. The process is the same as for #684 

We need to adapt the `ManagedWorkspace` model in CoreData and whatever repository model was created in #683 to work with sync, for that it needs:

- A local id (the one we already have) and a serverId. Check the types of these with @simonrozsival. Ideally these two should be type safe and different between each other so we don't confuse them, which could potentially happen easily. Use something like the GenericIdentifier we use currently in business models.
- The following syncable property: name

For the syncable property we need 3 values:
- current (of the type of that property)
- backup (of the type of that property)
- syncStatus (.inSync, .syncNeeded, .syncing)

There are multiple ways we could do this in CoreData, but the most performant one is probably to have those plain properties in the ManagedWorkspace model. In any case the specific approach should be decided by whoever takes this issue. Take into account that we could (should) map them to their own generic structs and enums when exiting the DB module.

## Dependencies
#683 

## Definition of done
- [ ] The name sync properties are added to the DB

Odrakir:
## Info
Adds the necessary code and changes to the DB to make Workspaces syncable.

### Summary
The changes to `ManagedWorkspace` consist in adding `serverId` and replacing `name` with `name_current`, `name_backup` and `name_syncStatus`, as described by the issue. A getter/setter is added to ease access to the name property.

For `WorkspaceDAO` I use a property wrapper to make access to the name property transparent when we just want the current value.

I've also added the `WorkspaceFeatures` that we were missing.

### Technical considerations
I left `serverId` as an `Int64` instead of a `GenericIdentifier` because I think it's more helpful to differentiate it from `id` than anything else. If we feel we need to make it type safe we would have to create a new type for it.

### Relationships
Closes #769 

### Pull Request checklist

### Pointers
Those are all very small commits and they should make sense by them selves.

### Review template

- [ ] The code considers i18n

## Merge permissions
Go!


Odrakir:
Since we already know those are needed by looking at android I also added `syncStatus` and `lastSyncErrorMessage` properties.

sofokli1 (commented):
Why does only the WS name need `current` and `backup` but all other properties do not and all other model names do not?

Odrakir:
`Name` is the only property on Workspace that's syncable. For other entities there'll be more, but this PR only changes Workspaces.

sofokli1 (approved):
looks good

PR #771, opened 2020-09-24
Settings: Implement Smart Alerts setting

### LINKED ISSUE #492: Settings: Implement Smart Alerts setting

sofokli1:
### Summary
Add the logic and UI for when the app does not have notification permissions. It works correctly and updates the subsetting for smartAlert based on notification permissions. Also updates the main settings view to show the selected smartAlertOption or nothing depending on notification permissions

### Relationships
Closes  #492

### Pull Request checklist

### Pointers
Commit by commit

### Review template
- [ ] The code considers i18n

## Merge permissions
🦑


Dependencies/Assets/Sources/Assets/Strings/en.lproj/Localizable.strings
```diff
@@ -138,3 +138,6 @@
 "badSSOEmailException" = "User email doesn't match SSO credentials";
 "invalidConfirmationCodeSSOException" = "Invalid or expired confirmation code";
 "singleSignOnError" = "Single Sign On is not configured for your email address. Please try a different login method or contact your administrator.";
+"Allow Access" = "Allow Access";
+"Open Settings App" = "Open Settings App";
+"smartAlertHeader" = "Notifications are currently disabled. If you'd like to receive notifications, please enable them in the Settings app.";
```
sofokli1, line 143: Ref #715

Dependencies/Settings/Sources/Settings/Common/View/SettingsTableViewController.swift
```diff
@@ -101,12 +103,30 @@ public class SettingsTableViewController: UIViewController, Storyboarded {
...
+                Driver.combineLatest(
+                    store.select(\.self),
+                    NotificationCenter.default.rx
+                        .notification(UIApplication.didBecomeActiveNotification)
+                        .map { _ in Void() }
+                        .asDriver(onErrorJustReturn: Void())
+                        .startWith(Void())
+                )
+                .flatMap { (state, _) in Driver.combineLatest(Driver.of(state), self.notificationsService.authorizationStatus().asDriver(onErrorJustReturn: .notDetermined)) }
+                .map(multiOptionSettingsSections)
```
sofokli1, line 129: @Odrakir this looks ugly but am open to suggestions

Dependencies/OtherServices/Sources/OtherServices/NotificationsService.swift
```diff
@@ -13,6 +14,17 @@ public class NotificationsService {
...
+    public func authorizationStatus() -> Single<UNAuthorizationStatus> {
+        return Single.create { single in
+            UNUserNotificationCenter
+                .current()
+                .getNotificationSettings { (settings) in
+                    single(.success(settings.authorizationStatus))
+                }
+            return Disposables.create { }
+        }
+    }
```
Odrakir, line 26: I think this one should also be dispatching an action to change the state. Maybe it can be a subscription (more or less the same way DB subscriptions work, take a look). I'll simplify a lot the code in the VC that selects the state.

Dependencies/TogglTrack/Sources/TogglTrack/App/Views/AppViewController.swift
```diff
@@ -12,13 +12,21 @@ class AppViewController: UINavigationController {
...
     private var disposeBag = DisposeBag()
 
-    init(store: Store<AppState, AppAction>, time: Time, signInService: SignInService, ssoService: SSOService) {
+    init(
+        store: Store<AppState, AppAction>,
+        time: Time,
+        signInService: SignInService,
+        ssoService: SSOService,
+        notificationsService: NotificationsService) {
+
```
Odrakir, line 24:
```suggestion
        notificationsService: NotificationsService
    ) {    
```

Odrakir: Although if you do the subscription thing I mentioned, you won't need to inject this service anymore.

Odrakir (dismissed):
Ping me if it's not clear.

Dependencies/OtherServices/Sources/OtherServices/NotificationsService.swift
sofokli1: ``getNotificationSettings` fires only once as far as I know. How will it dispatch every time?

Odrakir: Correct, my bad. Then this should probably be merged with `requestAuthorization` in order to get the current state of the authorization regardless if it was asked before... couldn't we do that? 🤔

sofokli1: we want to check if the user manually changed the notification permissions after those where granted to us, not just when we ask for permission.

Odrakir: But when do you need to check it? Is it when ever the app comes from background? Then we just have to move that check to a reducer and have that logic there.

sofokli1: moved it to redcer

Dependencies/TogglTrack/Sources/TogglTrack/TogglTrack.swift
```diff
@@ -46,7 +46,12 @@ public class TogglTrack {
         window.makeKeyAndVisible()                       
     }
 
-    public func appDidBecomeActive() {
+    public func sceneDidBecomeActive() {
         store.dispatch(.timer(.logSuggestion(.loadSuggestions)))
+        store.dispatch(.appLifecycleUpdated(.active))
```
Odrakir, line 51: loadSuggestions could be dispatched as an effect from appLifecycleUpdated now (hint, you can use `Effect.concatenate`)

App/App/SceneDelegate.swift
```diff
@@ -37,7 +37,7 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
     func sceneDidBecomeActive(_ scene: UIScene) {
         // Called when the scene has moved from an inactive state to an active state.
         // Use this method to restart any tasks that were paused (or not yet started) when the scene was inactive.
-        togglTrack.appDidBecomeActive()
+        togglTrack.sceneDidBecomeActive()
```
Odrakir, line 40: Is his called when the app starts for the first time? If not, we should do something about it.

Odrakir (dismissed):
This is much better. Just a couple of things.

App/App/SceneDelegate.swift
sofokli1: it is called on first start too

Odrakir (approved):

PR #772, opened 2020-09-25
Add grouping UI

### LINKED ISSUE #734: Add grouping UI
## Description
We need to add the UI to show grouped TEs in the timelog and also the ability to expand/collapse groups.

The logic is already there, even `TimeEntryCellViewModel` has all the info we need. All this is missing is the UI.

## Definition of done
- [ ] UI added to show the different types of cells (single, group header open, group header closed, in group)
- [ ] Actions binded for expanding/collapsing

dshr:
### Summary
This PR takes care of TE grouping

### Relationships
Closes #734

### Pull Request checklist

### Pointers
Commit by commit should do it

### Review template
- [ ] The code considers i18n

## Merge permissions
Go for it


Dependencies/Timer/Sources/Timer/TimeEntriesLog/Views/TimeEntriesLogViewController.swift
```diff
@@ -45,7 +46,7 @@ public class TimeEntriesLogViewController: UIViewController, Storyboarded {
         navigationItem.rightBarButtonItem = settingsBarButtonItem
     }
 
-    // swiftlint:disable function_body_length
+    // swiftlint:disable function_body_length cyclomatic_complexity
```
Odrakir, line 49: Maybe it's time to create a custom datasource subclass and put all the cell configuration code there.

Odrakir (changes requested):
Other than the comment: 
- Do we really need different cells for this? I see they are very similar, all they need is to be able to show/hide the number button (and potentially the stacked tes in the future). Maybe it's easier to do future updates if we use just once cell view.
- The tests that are failing are the ones `TimeEntriesSelectorTests`, since you changed how grouping works. Take a look.

dshr:
> Do we really need different cells for this? I see they are very similar, all they need is to be able to show/hide the number button (and potentially the stacked tes in the future). Maybe it's easier to do future updates if we use just once cell view.

@Odrakir It's not just the number button and stacked TEs, but the corners/sizing/bg colours as well. Ideally we can make all that configurable, but it kind of feels forced to me — that's going to be about 5 params each time we dequeue a cell.

As for the custom datasource — gonna do that, it does make sense!

PR #773, opened 2020-09-25
Make entities syncable

### LINKED ISSUE #685: Make other entities syncable
## Description
The same we did it for Time Entries in #684, we need to adapt the rest of the models in CoreData and whatever repository model was created in #683 to work with sync, for that they needs:

- A local id (the one we already have) and a serverId.
- The syncable properties and syncable relationships as defined in the Android implementatio

This has to be done for: workspaces, clients, projects, tasks, tags (user preferences happen in another issue)

## Dependencies
#684 

## Definition of done
- [ ] The rest of the sync properties and relationships are added to the DB

Odrakir:
## Info
It makes all entities syncable except for Workspaces (which was already done in a previous PR) and TimeEntry, which being so complex will be done in the next PR by itself.

Be aware. This PR breaks the app since there are many changes to the DB and we are not bothering with migration for now. You can completely delete the app and install again, but then it'll crash when you try to add a new entity because we are not setting the default workspace properly.

Another PR should be done to populate the DB with dummy data or just fix the default workspace bug.

### Technical considerations
It basically follows the same technique done for workspaces but for the rest of the entities, one commit at a time. It also includes changes to mappings from DTOs and tests.

**Note:** I'm thinking we should probably also move the user to the DB and include the preferences in the same object, as they do on Android, it'll help with sync since it'll use the same code as everything else. I'll add a note to do it.

### Relationships
Closes #685 

### Pull Request checklist

### Pointers
Commit history is good.

### Review template

- [ ] The code considers i18n

## Merge permissions
Go!


Dependencies/Repository/Sources/Repository/Models/Mapping/ProjectDTO+Mapping.swift
```diff
@@ -5,13 +5,14 @@ extension ProjectDTO {
...
-            workspaceId: self.workspaceId,
-            clientId: self.clientId
+            name: Syncable.inSync(self.name),
+            private: Syncable.inSync(self.isPrivate),
+            active: Syncable.inSync(self.isActive),
+            color: Syncable.inSync(self.color),
+            billable: Syncable.inSync(self.billable),
+            clientId: Syncable.inSync(self.clientId),
+            workspaceId: Syncable.inSync(self.workspaceId),
+            syncStatus: EntitySyncStatus.syncNeeded
```
sofokli1, line 15: why all other are marked as `inSync` but the sync status itself is `syncNeeded`?

sofokli1 (changes requested):
Overall looks good. Left 2 question for clarification.
We can't have the create/edit TE in any broken state since for this and the next 2 weeks is the main focus for development. The crash needs to be fixed before merging this

Dependencies/Database/Sources/Database/Models/ManagedObjects/ManagedClient.swift
```diff
@@ -1,21 +1,45 @@
...
+            name_syncStatus = newValue.syncStatus.rawValue
+        }
+    }
+
+    public var workspaceId: Syncable<WorkspaceDAO.Identifier> {
+        Syncable(
+            current: workspace.id.toIdentifier(),
+            backup: workspace_backup.toIdentifier(),
+            syncStatus: PropertySyncStatus(rawValue: workspace_syncStatus)!
+        )
```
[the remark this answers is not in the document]
Odrakir: Because to get the actual workspace I need the context, and I don't have access to it from here. Relationship are a bit more complicated to do than normal attributes.

Dependencies/Repository/Sources/Repository/Models/Mapping/ProjectDTO+Mapping.swift
Odrakir: I'm not 100% sure about this, but it'll probably be solved in the next sync issues: My idea is that all new properties are in sync by default, and the whole new entity itself is in needs sync state.

Odrakir:
@sofokli1 This is not the code that's broken. The crash when creating a TE happens if there's not workspace with the user default workspace ID in the DB. To go around it just add a workspace with the workspace id that's in "me.txt", and the app will work fine from then on.

You could add this in Repository.load():

```swift
let workspaces = try db.workspaces.getAll()
if workspaces.count == 0 {
    db.workspaces.insert(
        model: WorkspaceDAO(id: 4442916.toIdentifier(), name: "Default", admin: true, features: Set([])))
}
```

Dependencies/Repository/Sources/Repository/Models/Mapping/ProjectDTO+Mapping.swift
Odrakir: This is what android does, by the way

sofokli1 (approved):

PR #774, opened 2020-09-28
Dismiss start edit view properly

### LINKED ISSUE #756: Start/Edit TE: Checkmark button not working very well
## Description
The checkmark button that finalises the TE creation and updates is not working well. When creating a TE it does not dismiss the view.

[Notion](https://www.notion.so/toggl/Start-edit-time-entry-5783684cca064055b4512819087375bb) [Designs](https://toggl.invisionapp.com/overview/[iOS]-Rebranding-ckem8staozb0k01wrhgs8msyd/screens/21197370)

## Dependencies
None
## Definition of done
- [ ] Make sure the button works as intended

Odrakir:
## Info
Fixes the start/edit view dismissing properly when the check mark button is tapped.

### Summary
This was not part of the milestone, but I added since it's an annoying bug and it was easy to fix.

The fix it self was just 2 lines of code, but then I decided that when decorating a reducer, in tests we should be testing the decorated version, so we can test the complete flow. The 5 commits following the first one are just about that.

This is to be testing reducers independently, but I think this adds much more value to our tests. Now we are able to actually test that hitting done and creating a new TE results in the view being dismissed.

### Relationships
Closes #756 

### Pull Request checklist

### Pointers
Nothing to say, it's easy stuff.

### Review template

- [ ] The code considers i18n

## Merge permissions
Go!


sofokli1 (approved):
LGTM

PR #775, opened 2020-09-28
Implement Calendar Settings

### LINKED ISSUE #491: Settings: Implement Calendar Settings. Includes everything that can happen after tapping Calendar Settings
## Description
Check the old app and try to do the same. Talk with the calendar feature champion to coordinate technical decision regarding calendar service etc.

## Dependencies
## Definition of done
- [ ] Implement the logic
- [ ] implement the UI

sofokli1:
### Summary
Adds the calendar settings and covers situations like losing permissions from the user and getting them back, selecting calendars and keeping them in store.

### Relationships
Closes #491

### Pull Request checklist

### Pointers
commit by commit and practical tests. try to break it.

### Review template
- [ ] The code considers i18n

## Merge permissions
🦑


Odrakir (dismissed):
The calendar settings move around when toggled, check slack

Dependencies/Settings/Sources/Settings/Common/SettingsState.swift
```diff
@@ -38,21 +40,33 @@ public struct SettingsState: Equatable {
...
         self.local = local
         self.entities = entities
         self.user = user
         self.appLifecycle = appLifecycle
-        self.authorizationStatus = authorizationStatus
+        self.notificationsAuthorizationStatus = notificationsAuthorizationStatus
+        self.calendarOSAuthorizationStatus = calendarOSAuthorizationStatus
+        self.selectedCalendarIds = selectedCalendarIds
+        self.userCalendars = userCalendars
+        self.calendarIntegrationEnabled = calendarIntegrationEnabled
```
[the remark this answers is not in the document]
sofokli1: all of these are needed in the settings. how can I get the cal stuff from settings?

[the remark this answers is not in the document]
Odrakir: You just pass them in the state initializer just like you do now but instead of coming from `self`, they come from `calendarState`:
```swift
            return SettingsState(
                local: localSettingsState,
                entities: localTimerState.entities,
                user: user,
                appLifecycle: appLifecycle,
                notificationsAuthorizationStatus: notificationsAuthorizationStatus,
                calendarOSAuthorizationStatus: calendarState.calendarOSAuthorizationStatus,
                selectedCalendarIds: calendarState.selectedCalendarIds,
                userCalendars: calendarState.userCalendars,
                calendarIntegrationEnabled: calendarState.calendarIntegrationEnabled
            )
```

Dependencies/TogglTrack/Sources/TogglTrack/App/AppState.swift
```diff
@@ -10,13 +10,14 @@ import Utils
...
 import UserNotifications
+import EventKit
 
 public struct AppState {
 
     public var user: Loadable<User> = .nothing
     public var showSettings: Bool = false
     public var appLifecycle: AppLifecycle = .active
-    public var authorizationStatus: UNAuthorizationStatus = .notDetermined
+    public var notificationsAuthorizationStatus: UNAuthorizationStatus = .notDetermined
```
Odrakir, line 20: I think this one should be moved to Settings too.

Dependencies/TogglTrack/Sources/TogglTrack/App/AppState.swift
```diff
@@ -10,13 +10,14 @@ import Utils
 import CalendarService
 import Settings
 import UserNotifications
+import EventKit
```
Odrakir, line 13: 🔥

Dependencies/Settings/Sources/Settings/Common/Models/SettingsSectionViewModel.swift
```diff
@@ -3,81 +3,39 @@ import RxDataSources
...
-    case workspace([SettingRowType])
-    case profile([SettingRowType])
-    case dateAndTime([SettingRowType])
-    case timer([SettingRowType])
-    case calendar([SettingRowType])
-    case general([SettingRowType])
-    case signout([SettingRowType])
-    case subSettingSelection([SettingRowType])
+    case sectionWithoutTitle([SettingRowType])
+    case sectionWithTitle([SettingRowType], title: String)
```
Odrakir, line 7: Isn't it enough to have title as an optional String?

Dependencies/Settings/Sources/Settings/Common/SettingsSelector.swift
```diff
@@ -75,32 +77,58 @@ let multiOptionSettingsSections: (SettingsState) -> [SettingsSectionViewModel] =
...
-        sections.append(SettingsSectionViewModel(original: .subSettingSelection([]), items: rows))
+        sections.append(SettingsSectionViewModel(original: .sectionWithoutTitle([]), items: rows))
+    case .calendar:
+        var rows: [SettingRowType] = [ToggleRow(title: Strings.Settings.allowCalendarAccess, action: .allowCalendarAccessToggled, value: state.calendarIntegrationEnabled && state.calendarOSAuthorizationStatus == .authorized)]
+        if state.calendarOSAuthorizationStatus != .authorized {
+            rows.append(InfoRow(title: Strings.Settings.calendarSettingFooter, action: nil))
+        }
+        sections.append(SettingsSectionViewModel(original: .sectionWithoutTitle([]), items: rows))
+        if state.calendarIntegrationEnabled && state.calendarOSAuthorizationStatus == .authorized {
+            _ = state.userCalendars
```
Odrakir, line 113: This shouldn't be done like this, ignoring the result of this chain of methods and updating the sections var inside map is not good practice.

The chain of methods should result in the sections that need to be appended, and then we append those.

Odrakir (changes requested):

Dependencies/TogglTrack/Sources/TogglTrack/App/AppReducer.swift
```diff
@@ -36,13 +36,21 @@ func createAppReducer(environment: AppEnvironment) -> Reducer<AppState, AppActio
...
                     Effect.just(.timer(.logSuggestion(.loadSuggestions)))
                 )
             }
             return checkNotificationAuthorizationStatusEffect(environment.notificationsService)
 
-        case let .updateAuthorizationStatus(status):
-            state.authorizationStatus = status
+        case let .updateNotificationsAuthorizationStatus(status):
+            state.localSettingsState.notificationsAuthorizationStatus = status
+            return .none
```
Odrakir, line 47: This should be a `SettingsAction` now.

Dependencies/TogglTrack/Sources/TogglTrack/App/AppReducer.swift
```diff
@@ -36,13 +36,21 @@ func createAppReducer(environment: AppEnvironment) -> Reducer<AppState, AppActio
...
+
+        case .updateCalendarData:
+            state.localCalendarState.calendarOSAuthorizationStatus = environment.calendarService.getAuthorizationStatus()
+            state.localCalendarState.userCalendars = environment.calendarService.getAvailableCalendars()
+            state.localCalendarState.selectedCalendarIds = environment.userDefaultsService.selectedUserCalendarIds
+            state.localCalendarState.calendarIntegrationEnabled = environment.userDefaultsService.useCalendarIntegration
             return .none
```
Odrakir, line 54: This should be a `CalendarAction`

Dependencies/Settings/Sources/Settings/Common/SettingsSelector.swift
```diff
@@ -75,32 +77,58 @@ let multiOptionSettingsSections: (SettingsState) -> [SettingsSectionViewModel] =
...
+        if state.calendarIntegrationEnabled && state.calendarOSAuthorizationStatus == .authorized {
+            state.userCalendars
+                .grouped(by: \.sourceName)
+                .sorted(by: { (group1, group2) -> Bool in
+                    guard let group1Name = group1.first?.sourceName, let group2Name = group2.first?.sourceName
+                    else { return true }
+
+                    return group1Name < group2Name
+                })
+                .forEach { sourceGroup in
```
Odrakir, line 121: I personally prefer the more functional approach using `map`, but ok.

Dependencies/Settings/Sources/Settings/Common/Models/SettingsSectionViewModel.swift
```diff
@@ -3,84 +3,33 @@ import RxDataSources
 import Assets
 
 enum SettingsSectionViewModel: Equatable, SectionModelType {
```
Odrakir, line 5: But why is this still an enum now?

Odrakir (changes requested):

Dependencies/TogglTrack/Sources/TogglTrack/App/AppReducer.swift
sofokli1: the issues with this approach is that after moving things to the settings action/reducer it crashes because doing the following in the appReducer
```
case let .updateNotificationsAuthorizationStatus(status):
            return Effect.just(.settings(.updateNotificationsAuthorizationStatus(status)))
```
will fatal error with the below message since the settings settingState is optional and nil at the this point since the view is not showed yet.
```
Fatal error: "UpdateNotificationsAuthorizationStatus: UNAuthorizationStatus" was received by an optional reducer when its state was "nil". This can happen for a few reasons:
* The optional reducer was combined with or run from another reducer that set "SettingsState" to "nil" before the optional reducer ran. Combine or run optional reducers before reducers that can set their state to "nil". This ensures that optional reducers can handle their actions while their state is still non-"nil".
* An active effect emitted this action while state was "nil". Make sure that effects for this optional reducer are canceled when optional state is set to "nil".
* This action was sent to the store while state was "nil". Make sure that actions for this reducer can only be sent to a view store when state is non-"nil". In SwiftUI applications, use "IfLetStore".: file
```

What do you suggest here?

PR #777, opened 2020-09-28
Make TimeEntries Syncable

### LINKED ISSUE #684: Make time entries syncable
## Description
We need to adapt the `ManagedTimeEntry` model in CoreData and whatever repository model was created in #683 to work with sync, for that it needs:

- A local id (the one we already have) and a serverId. Check the types of these with @simonrozsival. Ideally these two should be type safe and different between each other so we don't confuse them, which could potentially happen easily. Use something like the GenericIdentifier we use currently in business models.
- The following syncable properties:  description, startTime, duration, billable, isDeleted
- The following syncable relationships: workspaceId, projectId, taskId

For the syncable properties we need 3 values for each of them:
- current (of the type of that property)
- backup (of the type of that property)
- syncStatus (.inSync, .syncNeeded, .syncing)

For the syncable relationships is the same with the caveat that while `current` will be the real CoreData relationship (mapped to an ID when exiting the DB module), `backup` on the other hand will always be the ID of the related entity. 

There are multiple ways we could do this in CoreData, but the most performant one is probably to have those plain properties in the TimeEntry model. In any case the specific approach should be decided by whoever takes this issue. Take into account that we could (should) map them to their own generic structs and enums when exiting the DB module.

## Dependencies
#683 

## Definition of done
- [ ] The rest of the sync properties and relationships are added to the DB

Odrakir:
## Info
The same way as it was done for the rest of entities, this adds the needed attributes to Time entries to make them syncable.

### Technical considerations
Not much to say, it follows the same exact technique as the rest.

### Relationships
Closes #684 

### Pull Request checklist

### Review template
It's pretty simple

- [ ] The code considers i18n

## Merge permissions
Go!


dshr (approved):
LGTM

